PR #2, opened 2015-01-27
Updates dependency from lodash-node 2.4.1 to lodash 3.0.0

martintietz:
[lodash-node seems to be deprecated](https://github.com/lodash/lodash-node). This pull request updates the dependency to use lodash 3.0.0 instead.


gaearon:
Great, thank you.

gaearon:
It's on NPM.

Issue #6, opened 2015-02-26
Add standalone UMD build with Webpack


gaearon:
This is up for grabs for anyone interested.
I'd like to export `Normalizr` as global.

Example Webpack config for a library: https://github.com/gaearon/react-dnd/blob/master/webpack.config.js

Once we do this we call it stable and release as 1.0!

gaearon:
We ship UMD build in `dist` folder of NPM package now.

PR #11, opened 2015-07-24
Feature/webpack

vlkosinov:
What do you think? :blush:
#6


gaearon:
Great! Let's use rimraf so it builds on Windows too?

vlkosinov:
@gaearon done :+1:

gaearon:
If we use babel we should also compile CommonJS to `lib` folder using Babel CLI and change `main` in package json to point to the compiled version.

gaearon:
Thank you!

tadjik1:
good job @vlkosinov !

gaearon:
Thanks again! This got me to finally release 1.0 :-P

PR #13, opened 2015-08-01
feat: update readme examples to ES6

theaqua:
Why not?


urbanvikingr:
@theaqua Eugene, do you mind and correct:
```
const normalized = normalize(response, {
  users: arrayOf(users)
});
```
to:
```
const normalized = normalize(response, {
  users: arrayOf(user)
});
```

theaqua:
@urbanvikingr right.

Issue #10, opened 2015-06-05
Duplicate entityKeys


jarno-steeman:
This is my Article structure 
```javascript
{
	Article: {
		id: 1,
		text: '...',
		AuthorId: 1 // -> User
	}
}
```
If I query my database, and include the Author relation, I'd end with something like this:
```javascript
{
	Article: {
		id: 1,
		text: '...',
		AuthorId: 1,
		Author: {
			id: 1,
			name: '...'
		}
	}
}
```
All peachy, but normalizr would turn something like that into this:
```javascript
{
	Article: {
		id: 1,
		text: '...',
		AuthorId: 1,
		Author: 1
	}
}
```
As my response objects grow, these extra fields strart to add up if you have allot of relations. 

If you'd be running with a document database like mongo you'll probably not have an AuthorId defined in the first place, but this is different for lets say a MySQL database. It would be awesome if there was an option to disable this piece of code:

```javascript
if (!bag[entityKey]) {
    bag[entityKey] = {};
}
```

gaearon:
What do you suggest the normalized output should look instead in your case?

jarno-steeman:
Without the Author key i'd say. If you work with some kind of models library you'd have the AuthorId field defined in there and use that anyway.

gaearon:
Are you running normalizr on server? Trying to understand the use case better..

jarno-steeman:
Ahh yes, I am. Forgot to mention that.

gaearon:
I'd accept a PR that lets you specify a third optional `options` argument to `normalize` call.
The only supported option would be `assignEntity`.

It default implementation is

```js
function assignEntity(obj, key, val) {
  obj[key] = val;
}
```

It would be called from [this line](https://github.com/gaearon/normalizr/blob/22c90e2e441444ae54a411c9aff137522ac8bc55/index.js#L13) instead of the direct assignment a la

```js
// before:
// normalized[prop] = visit(obj[prop], schema[prop], bag);

var entity = visit(obj[prop], schema[prop], bag);
assignEntity(normalized, prop], entity);
```

(and you'd need to drag `assignEntity` into that function from the main `normalize` call)

For your use case, you'd implement something like

```js
function assignEntity(obj, key, val) {
  obj[key] = val;
  delete obj[key + 'Id'];
}
```

If you do decide to PR this, please include tests.

Thanks!

jarno-steeman:
[...]

PR #14, opened 2015-08-05
Add assignEntity via options param to normalize

babsonmatt:
normalize method now takes a 3rd options param with only supported option being assignEntity

See issue #10

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -11,7 +11,10 @@ function visitObject(obj, schema, bag, assignEntity) {
   for (var prop in obj) {
     if (obj.hasOwnProperty(prop)) {
       var entity = visit(obj[prop], schema[prop], bag);
-      assignEntity(normalized, prop, entity);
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -5,24 +5,28 @@ var EntitySchema = require('./EntitySchema'),
...
+function visitObject(obj, schema, bag, options) {
   var normalized = {};
 
   for (var prop in obj) {
     if (obj.hasOwnProperty(prop)) {
-      normalized[prop] = visit(obj[prop], schema[prop], bag);
+      var entity = visit(obj[prop], schema[prop], bag, options);
+      if (options && options.hasOwnProperty('assignEntity'))
+        options.assignEntity(normalized, prop, entity);
+      else
```
gaearon: Can we instead have default options with default `assignEntity` implementation?

Issue #17, opened 2015-08-18
denormalizr?


tlrobinson:
(I'm new to redux/normalizr so tell me if I'm totally off-base)

It's often more convenient to pass nested objects down as props, so I find myself writing selectors like so:

```
    return {
        foo: {
            ...foo,
            bars: foo.bars.map(id => state.bars[id])
        }
    };
```

Am I doing it wrong, or would it be nice to have a `denormalizr` that converts the normalized objects back to nested objects (perhaps using the same schema definition?)

dotcs:
I am working on this. As suggested it is entirely possible to use the same schema definition to convert back to nested objects. I will add a PR within the next days to see what you guys think about it. Stay tuned.

gaearon:
Closing, as this is duplicate of https://github.com/gaearon/normalizr/issues/1.
See also the discussion there.

PR #20, opened 2015-08-24
Denormalization with the same schema that is used for normalization

dotcs:
My attempt to revert the normalisation by using the same schema. This was suggested by #17 and #1. I am not sure what you think about this @gaearon but I'd love to hear your feedback on this.

Note also that I have inverted all of the relevant test cases, except these two:

* can normalise nested entity and delete an existing key using custom function
* can normalise mutually recursive entities

I think the first one could be archived by using the options param, that is new in v1.0.0. But I don't really see a use case here. The latter one is difficult because of the recursive entities. If you have a good idea how to break the dependency chain, feel free to give me some feedback on this. This is also the reason why I have written the single test for the denormalisation but also disabled it for now.

One more thing: in principle one could just define the bag once, in the `denormalize` call, and access the bag from within all functions that need access, but I kept the same function arguments as in the normalisation direction. I think this is purely a matter of taste.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -20,6 +20,15 @@ function visitObject(obj, schema, bag, options) {
   return normalized;
 }
 
+function visitObjectForDenorm(obj, schema, bag) {
+  var denormalized = {},
+    key = Object.keys(obj)[0];
+
+  denormalized[key] = visitForDenorm(obj[key], schema[key], bag);
+
+  return denormalized;
+}
+
 function visitArray(obj, arraySchema, bag, options) {
   const itemSchema = arraySchema.getItemSchema();
 
@@ -29,6 +38,30 @@ function visitArray(obj, arraySchema, bag, options) {
   return normalized;
 }
 
+function visitArrayForDenorm(obj, arraySchema, bag) {
+  var itemSchema = arraySchema.getItemSchema(),
+    itemSchemaKey = itemSchema.getKey(),
+    denormalized;
+
+  var item = bag[itemSchemaKey];
+
+  denormalized = [];
+  obj.forEach(function(itemKey) {
+    var keys;
+    if (item.hasOwnProperty(itemSchemaKey)) {
+      denormalized.push(visitForDenorm(item[itemKey], itemSchema[itemSchemaKey], bag));
+    } else {
+      keys = Object.keys(itemSchema);
+      if (keys.length <= 2) {
+        denormalized.push(visitForDenorm(item[itemKey], undefined));
+      } else {
+        denormalized.push(visitForDenorm(itemKey, itemSchema, bag));
+      }
+    }
+  });
+
+  return denormalized;
+}
 
 function mergeIntoEntity(entityA, entityB, entityKey) {
   for (let key in entityB) {
@@ -68,6 +101,19 @@ function visitEntity(entity, entitySchema, bag, options) {
   return id;
 }
 
+function visitEntityForDenorm(entity, entitySchema, bag) {
+  var entityKey = entitySchema.getKey();
+  var denormalized = bag[entityKey][entity];
+  Object.keys(entitySchema).forEach(function(schemaKey) {
+    if (schemaKey.indexOf('_') !== 0) {  // TODO: better way to access the relevant schema keys?
+      if (typeof denormalized === 'object' && denormalized.hasOwnProperty(schemaKey)) {
+        denormalized[schemaKey] = visitForDenorm(denormalized[schemaKey], entitySchema[schemaKey], bag);
+      }
+    }
+  });
+  return denormalized;
+}
+
 function visit(obj, schema, bag, options) {
   if (!isObject(obj) || !isObject(schema)) {
     return obj;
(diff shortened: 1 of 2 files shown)


rovolution:
+1 for a denormalization method

nhagen:
I think that denormalizing mutually recursive entities would be useful. Can you elaborate more on the problem there?

dotcs:
In this PR, the test "can denormalize mutually recursive entities" does fail because of recursive patterns. To circumvent the problem, I think some some user input is needed that defines when to stop a possible recursion (some upper boundary). Let me go into the details.

Think of the following schema: `USER {articles: [ARTICLE]}`, `ARTICLE {authors: [USER]}`, where the capitalized words are a schema each. You can easily create a nested entity, where a user has written a article and of cause is then the author of the article, and so on. This dependency chain can be nested arbitrarily deep like here.

```javascript
users: [
  {id: 1, username: 'Testuser', articles: [
    {id: 1, title: 'Testarticle', authors: [
      {id: 1, username: 'Testuser', articles: [
        id: 1, title: 'Testarticle', authors: [
          // ...
        ]
      ]}
    ]}
  ]}
]
```

Let's say we include only objects up to two levels, than we have:

```javascript
users: [
  {id: 1, username: 'Testuser', articles: [
    {id: 1, title: 'Testarticle', authors: [
      {id: 1, username: 'Testuser'}
      ]}
    ]}
  ]}
]
```

Please note, that I have removed the `articles` property on the second user object.

Now, when we use normalizr, we should get

```javascript
{
  result: {
    users: [1]
  },
  entities: {
    articles: {
      1: { id: 1, username: 'Testuser', articles: [1]}
    },
    users: {
      1: {id: 1, title: 'Testarticle', authors: [1]}
    }
  }
}
```

But doing the process in reverse does only work, when we specify how deep the result should be nested. It is the same process as above to avoid arbitrary deep nested objects. Because this is currently not done, the reverse action would try to nest the objects to an infinite level, which is impossible. Because of this the test fails.

Please note, that the test "can denormalize self-recursive entities" does pass, because the input is given in a form, such that it is clear where to stop the dependency chain. Note, that `parent` is missing on the user entity with ID 4 in this case.

```javascript
    input = {
      result: 1,
      entities: {
        users: {
          1: {
            id: 1,
            name: 'Andy Warhol',
            parent: 7
          },
          7: {
            id: 7,
            name: 'Tom Dale',
            parent: 4
          },
          4: {
            id: 4,
            name: 'Pete Hunt'
          }
        }
      }
    };
```

But of course this "trick" does not solve the problem above, where the input simply does not allow such a walkaround.

markerikson:
This looks useful.  Are there any plans to merge this in?

zuk:
Would like to see this merged in too.

zuk:
Shallow cloning the values pulled from `bag` (i.e. the state) seems to be enough to prevent mutation of the state:

```
import clone from 'lodash/lang/clone'
...
  var item = clone(bag[itemSchemaKey])
...
  var denormalized = clone(bag[entityKey][entity])
```

lucasfeliciano:
And about many to many relations?

How will you denormalize it?

Lets say we have:

`Products * -- * vendors`
`vendors 1  -- * offers`

The algorithm will not be able to put the vendors objects in the right parent, if we have the same vendor attributed to 2 different products.

gaearon:
I don't think we want to support denormalization in this project.
It is likely to have bad performance and many edge cases.

However I'm happy to add some “official” schema access methods so that you can build a denormalizr as a separate project without relying on internal details. Please feel free to propose such API in a separate PR.

mdgbayly:
@dotcs Did you ever pursue moving your forked code with denormalization support to another package?

I'm surprised at how little support there is for denormalization.  Maybe I'm missing something....
We load data from an API and we're moving towards using normalizr to make it easier to handle the data in our redux app.

However, after we've updated the data in our app we have to be able to save it back to the API again in the original shape. Hence the need for something like denormalization.

I recognize it might be complicated. But how else are people saving data back to an API after they've normalized it?  Or are they only saving simple entities back?  Maybe our API format is too complex, and it's a smell that we're trying to save data that needs to be denormalized.

But anyway, I've taken a local copy of your denormalize function and tried it out on our relatively simple use cases and it seems to support what we need (so far only nested arrayOf type schemas).

zuk:
We ended up forking the denormalizr code and expanding on it quite a bit. Been meaning to post the code back to GitHub. Sounds like there's some interest for this sort of thing... I'll try to get it posted this week.

mschipperheyn:
Here's a simple marshaller for those who are interested: https://gist.github.com/mschipperheyn/a612dc7e15a5ff28fc91932be2de09eb

mschipperheyn:
I initially talked about implementing a proxy here https://github.com/gaearon/normalizr/issues/78 . But at that time, it didn't seem to make a lot of sense, since there doesn't seem to be a real performance gain from proxying the object in stead of marshalling except if the objects are very large. But then we started using immutable on our store. And suddenly, we find ourselves converting stuff to POJOs all the time in order to marshal them.

Now, creating a proxy seems to make a lot more sense. If we normalize an incoming json and store it in the store as an immutable object, getting the associated entity that is identified by id, could simply return the immutable entity that already exists in the "entity store".

I'm not an expert on javascript proxies (more familiar with java), but what if, instead of normalizing a user object to

user: 1 

we would do this instead?

user:function(id = 1){
   return userEntitiesReference[id];
}

Any thoughts?

YoruNoHikage:
I had a thought on this (for https://github.com/gpbl/denormalizr) about denormalizing with getters, it's an awesome feature in javascript where you can transparently have a function for when you want to access a specific property. More info about [lazy getters on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get#Smart_self-overwriting_lazy_getters)

So when you want to access `user.relatedProperty`, it actually calls for denormalizing this property (setting lazy getters on the related property).

paularmstrong:
So, I'm going to eat my hat. #214

PR #22, opened 2015-09-17
Include an implementation of isEqual and isObject

serhat-xyz:
Minified version of lodash is 48.7 K, and the only functions needed in normalizr are `isEqual` and `isObject`. This patch removes that dependency.

`normalizr.min.js` size:

```
before: 8.7 K
after:  3.4 K
```

Please let me know what you think about this patch, any feedback is greatly appreciated :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,10 +388,6 @@ AppDispatcher.register((payload) => {
 });
 ```
 
-## Dependencies
-
-* `lodash` for `isObject` and `isEqual`
-
 ## Running Tests
 
 ```
--- a/package.json
+++ b/package.json
@@ -34,8 +34,5 @@
     "mocha": "^2.2.5",
     "rimraf": "^2.4.2",
     "webpack": "^1.10.5"
-  },
-  "dependencies": {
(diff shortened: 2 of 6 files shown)


unindented:
@serhat-xyz The code is loading only the functions that it needs (e.g. `require('lodash/lang/isObject')`). When functions are required that way, only the necessary parts of lodash are loaded, not the whole library, as long as you use something like browserify or webpack.

Also, all functions in lodash are published independently (https://www.npmjs.com/browse/keyword/lodash-modularized), so if I were you I'd require [`lodash.isequal`](https://www.npmjs.com/package/lodash.isequal) and [`lodash.isobject`](https://www.npmjs.com/package/lodash.isobject) instead of including your custom implementations.

gaearon:
I would like to avoid maintaining these functions as part of normalizr.
Rather, I'd wait for Lodash 4 and do something similar to https://github.com/rackt/redux/pull/611.

serhat-xyz:
Thanks for the feedback, I learned some stuff with this :)

PR #24, opened 2015-10-05
Add ability to specify custom id attribute function

bitantics:
Hello!

I have ran into a few cases where a server -- not controlled by me -- returns nested JSON where:
* there isn't a single key which can serve as an identifier slug
* or the slug is embedded further within the object
* or the key requires some simple preprocessing

This PR uses the existing API of normalizr, but extend the `idAttribute` option to accept arbitrary functions. An example use case is in the included test.

### DIFF
--- a/src/EntitySchema.js
+++ b/src/EntitySchema.js
@@ -4,16 +4,18 @@ export default class EntitySchema {
       throw new Error('A string non-empty key is required');
     }
 
-    this._idAttribute = options.idAttribute || 'id';
     this._key = key;
+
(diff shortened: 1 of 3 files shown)


woodb:
:+1: So awesome, I was just about to make this PR myself when I was looking for this feature :)

gaearon:
Great PR.
:+1:

gaearon:
Out in 1.1.0

bitantics:
Thanks for the quick response, it's much appreciated!

Here's a dolphin: :dolphin:

PR #29, opened 2015-10-16
Allow visiting map values

### LINKED ISSUE #28: Dealing with maps
I'm probably working with the weirdest payload ever, but it's a legacy API, so bear with me...

The beast looks something like this:

```js
{
  "messages": [{
    "id": 1,
    "body": "A message",
    "attachments": [{
      "id": 1,
      "type": "images",
      ...
    }]
  }],
  "replies": {
    "1": [{
      "id": 2,
      "body": "First reply",
      "attachments": [{
        "id": 2,
        "type": "images",
        ...
      }, {
        "id": 3,
        "type": "videos",
        ...
      }]
    }]
  }
}
```

So `messages` is an array of thread starters / OPs, and `replies` is a hash where the keys are message IDs, and the values are arrays of replies to the corresponding message.

I'd like to normalize both messages and replies as `message` entities, but I don't have a way of traversing the `replies` hash.

I was thinking of implementing a `mapOf` operator that would visit the values of a hash. The solution to my problem would then look like:

```js
var message = new Schema('messages')
var attachment = {
  images: new Schema('images'),
  videos: new Schema('videos')
}

message.define({
  attachments: arrayOf(attachment, {schemaAttribute: 'type'})
})

normalize(json, {
  messages: arrayOf(message),
  replies: mapOf(arrayOf(message)),
}
```

Does that sound useful?

unindented:
Fixes #28.

@gaearon Two points on which I'd like to hear your opinion:

* I reused `ArraySchema` to implement this functionality, as it is really similar to `arrayOf`. I decided to rename `ArraySchema` to `IterableSchema` though, as it is no longer used only for arrays.
* I imported `mapValues` from lodash to deal with transforming the map values. I could have avoided the import, but I thought it was clearer this way.

Cheers!


gaearon:
This looks good to me.
Anything pending, or shall I merge and release?

unindented:
Ready to go.

PR #35, opened 2015-10-28
Custom `mergeIntoEntity`

### LINKED ISSUE #34: Custom `mergeIntoEntity`
I have this scenario where an entity appears twice in a payload, but with slightly different attributes. The problem is that some of those attributes are deeply nested, so the default `mergeIntoEntity` strategy doesn't cut it, as I would lose data:

```js
{
  foo: [{
    id: 1,
    contact: {
      email: '...'
    }
  }],
  bar: [{
    id: 1,
    contact: {
      telephone: '...',
      email: '...'
    }
  }]
}
```

I was thinking of adding a `mergeIntoEntity` option, in the same line as the `assignEntity` option that we currently have, so that I can solve the conflict with my own code.

@gaearon does that sound reasonable?

unindented:
Fixes #34.

@gaearon what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,7 +330,11 @@ article.define({
 Normalizes object according to schema.  
 Passed `schema` should be a nested object reflecting the structure of API response.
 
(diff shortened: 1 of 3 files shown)


gaearon:
:+1:

gaearon:
Out in 1.4.0, thank you.

Issue #40, opened 2015-11-25
Different entity on single key


tobiaskun:
In this example `reference` can either be of type `shop_article` or `shop_outfit`. The function `valuesOf` treats the content as array and is looping over the keys.

```javascript
let reference = wish.define(
    shop_article: article,
    shop_outfit: outfit 
);

let example1 = {
    id: 1,
    _type: 'user_userhaswish',
    reference: {
        id: 1,
        type: 'shop_article'
    }
}

let example2 = {
    id: 1,
    _type: 'user_userhaswish',
    reference: {
        id: 1,
        type: 'shop_outfit'
    }
}
```

Is there a function like `valuesOf` that only expects one entity in the json?

gaearon:
Should be out in 2.0.0 via the new `unionOf()` (#42).

PR #42, opened 2015-12-11
unionOf -- extend polymorphic unions to fields

eyston:
For collection fields `arrayOf` and `valuesOf` both allow polymorphic schemas if you specify a `schemaAttribute` option.  I didn't see the ability to do this for a non-collection field.  The use case for this is if a field can also be polymorphic.

Mentioned in issue : https://github.com/gaearon/normalizr/issues/40

An example would be a `group` where the owner could be a `group` or a `user`.

```js
    var user = new Schema('users'),
        group = new Schema('groups'),
        owner = unionOf({
          users: user,
          groups: group
        }, { schemaAttribute: 'type' });

    group.define({
      owner: owner,
    });
```

This pull request adds the syntax `unionOf`.  It mimics the behavior found in `IterableSchema` but applies it in a non-iterable manner.

This could also pull the polymorphic logic out of `IterableSchema` by making `itemSchema` an instance of `UnionSchema` (the added type) if `schemaAttribute` is specified.

```js

arrayOf(unionOf({ groups: group, users: user}, { schemaAttribute: 'type' })) === arrayOf({ groups: group, users: user }, { schemaAttribute: 'type' })
```

This is my fist pull request so let me know if I'm doing this wrong!  Thanks :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,6 +325,52 @@ article.define({
 });
 ```
 
+### `unionOf(schemaMap, [options])`
+
+Describe a schema which is a union of multiple schemas.  This is useful if you need the polymorphic behavior provided by `arrayOf` or `valuesOf` but for non-collection fields.
+
+Use the required `schemaAttribute` option to specify which schema to use for each entity.
+
+```javascript
+const group = new Schema('groups');
+const user = new Schema('users');
+
+// a member can be either a user or a group
+const member = {
+  users: user,
+  groups: group
+};
+
+// You can specify the name of the attribute that determines the schema
+group.define({
+  owner: unionOf(member, { schemaAttribute: 'type' })
+});
+
+// Or you can specify a function to infer it
+function inferSchema(entity) { /* ... */ }
+group.define({
+  creator: unionOf(member, { schemaAttribute: inferSchema })
+});
+```
+
+A `unionOf` schema can also be combined with `arrayOf` and `valueOf` with the same behavior as each supplied with the `schemaAttribute` option.
+
+```javascript
+const group = new Schema('groups');
+const user = new Schema('users');
+
+const member = unionOf({
+  users: user,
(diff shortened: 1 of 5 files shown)


eyston:
the commit which removes polymorphic functionality from `IterableSchema` / `visitIterable` looks like:

https://github.com/eyston/normalizr/commit/99ff5c44c23e3126ec6c4faa0f1d862352efc23b

can merge that into this PR too if desired

gaearon:
I'm not currently actively using this module so it's a bit difficult for me to review.
@unindented What do you think?

gaearon:
This looks good to me on a glance. I'm happy to have eyston@99ff5c4 here too. Mind merging it here?

eyston:
merged https://github.com/eyston/normalizr/commit/99ff5c44c23e3126ec6c4faa0f1d862352efc23b

gaearon:
Can you please add it to the API docs in README and write release notes for the version bump?

eyston:
Yup, I didn't want to write docs until after any feedback ;p

unindented:
Looks good to me.

gaearon:
Then I'm ready to publish after docs are added!

eyston:
Thanks for your patience, I spent the last several days in a rabbit hole so put this off ;p.

I've added docs to the README but I'm not sure where to put the release notes.

gaearon:
Just post them here as a comment and I'll copy and paste.

eyston:
I found https://github.com/gaearon/normalizr/releases ;p

Here is my attempt!

- Adds support for polymorphic fields via `unionOf(schemaMap, options)` which works like the polymorphic support for `arrayOf` and `valuesOf` but for non-collection fields (#40).

gaearon:
Out in 2.0.0!

PR #43, opened 2015-12-14
Remove .babelrc from NPM-Package to support RN > v0.16.0

Purii:
In order to use normalizr with **react-native**, the **.babelrc** shouldn't be in the NPM-Package.
A "files"-whitelist in package.json, like gaearon/react-redux#206, would also be possible.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,1 +1,2 @@
 test
+.babelrc


mehdichamouma:
+1

chentsulin:
cc @gaearon

josebalius:
@gaearon I know you are busy these days :) anyway we can merge this?

gaearon:
Should be fixed in 1.4.1.

PR #46, opened 2016-01-05
Entity semantics, association chaining, sugar methods

smargh:
This PR includes a few different things that I considered splitting into different PRs, but I felt they were all conceptually inter-connected that making it a "one and done" made sense. Let me explain each change:

First, I've changed (with backwards compatibility) the semantics. Instead of exposing the `EntitySchema` as a `Schema` constructor, I've called it `Entity`. This makes much more semantic sense to my mind, as a "schema" is a graph of associations between entities. Thus, the standard usage of Normalizr would have someone define one "schema", made up of N "entities" each with N associations/relations to the other entities. What you are building by `define`ing relationships between "entities" is the "schema", and it is this full "schema" that allows Normalizr to parse various JSON responses. By calling the "entities" a `Schema`, you are mixing semantics, and I for one found it rather confusing. This PR adds an `Entity` alias for the `EntitySchema` constructor, but also leaves the `Schema` alias. So, it is fully backwards compatible, tho I have edited the README to use only the `Entity` alias.

Second, I changed the `.define()` method to allow for chaining (by `return`ing `this`). This too should be fully backwards compatible, but is necessary for the other changes I have made.

Namely, I have added two convenience methods: `.hasOne()` and `.hasMany()`. These function essentially as syntactic sugar for the two most common types of `.define()` calls, while also allowing for a cleaner semantic API. Because these methods simply call down to `.define()`, and `.define()` now allows for chaining, you can chain calls on a single entity, thus allowing for very clean code (IMO):

```javascript
const article = new Entity('articles');
const user = new Entity('users');
const collection = new Entity('collections');

article
  .hasOne(user, 'author')
  .hasMany(collection);

collection
  .hasOne(user, 'curator');
```

So, all in all, this PR adds very little new functionality, but it has made the way that I and my team use Normalizr so much easier, more direct, and more readable.

*Note*: I have not updated the tests yet. I will do so, however, if and when you say that this is worth pursuing.

stephen


gaearon:
@unindented Would you like to review this?

unindented:
@gaearon Yup, although it may have to wait until the weekend, as I'm having busy days at the office. Is that cool with you?

gaearon:
No problem. I'm too busy to review at all, so review in a week or two is infinitely better than no review :-)

Thank you.
(Please let me know if it becomes a burden!)

unindented:
Will do, but it isn't for now. Thank you for asking though.

gaearon:
To be honest I'm not a fan of "fluent APIs" because they obscure inputs and outputs.

I'm also not sure I'd want to see any shortcuts for `define`. While the current naming may not be perfect, I want to strongly make it clear that there is just a single way to put something in the schema, and that it can have an arbitrary shape, including nested objects:

```js
author.define({
  posts: arrayOf(post),
  position: {
    x: coordinate,
    y: coordinate
  }
})
```

Adding shortcuts like `hasOne()` and `hasMany()` makes this much harder to find, as most people will stay within the realm of those shortcuts, and will find it hard to think beyond them. This is why I'm inclined not to take this.

Thanks for the PR though, and I truly appreciate the effort. I just don't think the pros outweigh the cons in this particular case, and I don't want to introduce API churn as the library has been stable for quite some time, and I don't see enough reasons to justify changing things.

unindented:
@smargh Maybe that syntactic sugar can live in your own app, or you can publish a library with that simplified syntax that uses `normalizr` under the hood?

smargh:
Yeah. That's what I think I'll do. I have now have methods for `hasOne`, `hasMany`, `belongsTo`, and `hasAndBelongsToMany`. Plus, I added auto-bidirectional association mapping (for `hasAndBelongsToMany` associations) so that two entities that are bidirectionally associations each have an array of id keys. And then I added object templating/seeding for entity objects. So, I now have enough changes from the original library that I may just move forward with my own fork.

Thanks for the awesome library tho! This has proven absolutely essential to the app our team is working on.

PR #53, opened 2016-01-18
Make normalizr also copy Symbol keys

dlmanning:
So in the application I work on, we use normalizr both to slice up entities from the initial server response, and also internally to prepare nested entities constructed within forms to be merged into the entity branch of the state tree. We also use a Symbol key on our data objects in order to dispatch based on categorization at various places within the app. At present I need to "retag" objects with their appropriate category symbol key after they pass through normalizr. It would be nice if normalizr knew how to just copy Symbol keys.

I've made a couple of small changes to enable this that I don't *think* should present any significant downsides. I did use Object.keys to get the objects own string keys. I'm not sure if you were going for ES3 backward compatibility there, but it wouldn't be hard to replace with a `for... in` if you'd rather.

Sorry that the test I added is a bit weird. Mocha apparently only knows how to check for string keyed props at the moment, and so needs a little hand-holding to check for Symbols.

thoughts, feelings?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,6 +5,14 @@ import isObject from 'lodash/isObject';
 import isEqual from 'lodash/isEqual';
 import mapValues from 'lodash/mapValues';
 
+function getAllOwnPropertyKeys (obj) {
+  return Object.keys(obj).concat(
+    typeof Object.getOwnPropertySymbols === 'function'
+      ? Object.getOwnPropertySymbols(obj)
+      : []
+  )
+}
+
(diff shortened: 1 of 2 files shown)


marcodejongh:
@dlmanning This would also make Symbol work as an ID right?
I'm using normalizr for a json api that also has anonymous users which dont have a id, so figured the best way to store those is to give them a symbol as ID, but currently that doesn't work.

I'll give your code a go.

paularmstrong:
Closing due to no update from author in 6 months.

PR #56, opened 2016-01-20
Schema.define could return associated instance

xuoe:
Found myself expecting `Schema.define()` to return the Schema instance when doing:

```javascript
// In './Group.js'
import { arrayOf } from 'normalizr';
import User from './User';

export default new Schema('groups').define({
  users: arrayOf(User)
});
```

Thought it'd be a good idea if it did return it.

### DIFF
--- a/src/EntitySchema.js
+++ b/src/EntitySchema.js
@@ -29,5 +29,6 @@ export default class EntitySchema {
         this[key] = nestedSchema[key];
       }
     }
+    return this;
   }
 }


unindented:
Yup, I think this is fine.

unindented:
@xuoe could you add a unit test that checks that the instance is returned?

unindented:
Nevermind, I'll add it myself.

unindented:
@gaearon You may want to revert this one too if you think fluent APIs don't make sense here.

gaearon:
Yeah, sorry about the churn. We want to be very clear about which methods are mutating.

PR #60, opened 2016-02-01
Edit Usage and normalize() examples to explain second parameter better

fredrikolovsson:
This concerns improving the Usage example to better show how the second parameter for normalize() is used. 

Currently, the example in the Usage section has these two normalize calls, which handle responses with similar structure:
```
response = normalize(response, {
  articles: arrayOf(article)
});

response = normalize(response, {
  users: arrayOf(user)
});
```
The example could be more informative, if it showed how to normalize two different response object structures, like this:
```
// response is just one article object
response = normalize(response, article);

// response is an object with a reference to an array of article objects
response = normalize(response, {
  articles: arrayOf(article)
});
```
In addition to revising the Usage example, I suggest adding these examples to the normalize() specification.

It took me a while to figure out why my normalized data did not look as expected, and while I am new to the field, maybe someone else could also be saved some time by this, without making the README significantly longer.

Thanks for your great work and please let me know if there is something I should fix to make this a proper improvement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ Then we define nesting rules:
 ```javascript
 article.define({
   author: user,
+  contributors: arrayOf(user)
 });
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 # normalizr [![build status](https://img.shields.io/travis/gaearon/normalizr/master.svg?style=flat-square)](https://travis-ci.org/gaearon/normalizr) [![npm version](https://img.shields.io/npm/v/normalizr.svg?style=flat-square)](https://www.npmjs.com/package/normalizr) [![npm downloads](https://img.shields.io/npm/dm/normalizr.svg?style=flat-square)](https://www.npmjs.com/package/normalizr)
 
-Normalizes deeply nested JSON API responses according to a schema for [Flux](https://facebook.github.io/flux) and [Redux](http://rackt.github.io/redux) apps.
```
paularmstrong: Can you revert the trailing whitespace changes? In Markdown, the force an extra newline and this change loses them.

fredrikolovsson: Oh, I apologize for the beginner mistake. Thanks for letting me know. Should be fixed now: https://github.com/gaearon/normalizr/pull/60/commits/a74def24d9802a634cf6275ea073ba71d1962aa0

paularmstrong:
I agree that removing some of the keys that aren't fully explained helps. Your updates look good to me. Thanks!

PR #63, opened 2016-02-04
Remove unnecessary Array.isArray() checks

puradox:
The Array.isArray() checks included in the normalize function were not needed since lodash's isObject() already checks for valid arrays
```
_.isObject(value)
Checks if value is the language type of Object. 
(e.g. arrays, functions, objects, regexes, new Number(0), and new String(''))
```
[Source] (https://devdocs.io/lodash~4/index#isObject)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -120,11 +120,11 @@ export function unionOf(schema, options) {
 export { EntitySchema as Schema };
 
 export function normalize(obj, schema, options = {}) {
-  if (!isObject(obj) && !Array.isArray(obj)) {
+  if (!isObject(obj)) {
     throw new Error('Normalize accepts an object or an array as its input.');
   }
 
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -120,11 +120,11 @@ export function unionOf(schema, options) {
 export { EntitySchema as Schema };
 
 export function normalize(obj, schema, options = {}) {
-  if (!isObject(obj) && !Array.isArray(obj)) {
```
gaearon: I think you want to keep the first check: it is specifically whitelisting arrays.
The second check can probably be removed. Note that the first one has `!` but the other doesn't.

paularmstrong:
@puradox are you available to fix the noted issue and ensure tests pass?

puradox:
@paularmstrong I got confused with the comment by gaearon, but I think I made the necessary changes. It should be good now.

Issue #70, opened 2016-02-19
Keep consistency for relations


YoruNoHikage:
I think it would be great to have consistency in relationship between entities when normalized.
Here's an example : 

Consider you have two entities **user** and **project**. You define the relations like this (bidirectional) : 

``` js
userSchema.define({
  projects: arrayOf(projectSchema),
})

projectSchema.define({
  contributors: arrayOf(userSchema),
})
```

Now you want to create a project, you send some request to a server that replies with the project entity like this : 

``` js
{
  id: 80,
  contributors: [1], // your userId for example
  // other fields
}
```

You normalize the response and merge with the rest of your entities (just like the real-world example for Redux), the projects' entities are updated but not the user.

So my proposal is to add a capability of telling what is the property in the foreign object and instead of normalizing like this : 

``` js
{
  entities: {
    projects: {
      80: {
        id: 80,
        contributors: [1],
        // other fields
      }
    }
  }
}
```

I'm proposing something like this : 

``` js
{
  entities: {
    projects: {
      80: {
        id: 80,
        contributors: [1],
        // other fields
      }
    },
    users: {
      1: {
         id: 1,
         projects: [80]
       }
    }
  }
}
```

We could do something like this : 

``` js
userSchema.define({
  projects: arrayOf(projectSchema).mappedBy('contributors'),
})

projectSchema.define({
  contributors: arrayOf(userSchema).mappedBy('projects'),
})
```

This prevent any flaws in consistency while being simple (user side). I already tried a solution, it's not bulletproof but I can submit a cleaner PR if you're interested.

tokenvolt:
@YoruNoHikage I was thinking about the same feature, this is really would be very convenient.

lukekarrys:
I found myself needing this as well. Just to test I dropped in @YoruNoHikage's fork in my project and it worked exactly how I wanted it to for my simple case (which was pretty similar to users+projects). I think this would be a great PR.

MicheleBertoli:
@YoruNoHikage we started using your fork too (great job!), you should definitely submit a PR please.

YoruNoHikage:
Thanks, I'll clean the code a bit before submitting it. ;)

bpartridge:
I'd love to see something like this reach maturity! 

[...]

PR #80, opened 2016-03-28
Consistent relations, see #70

YoruNoHikage:
I already created an issue (see #70) to discuss consistency between relations but people are asking me to create a PR, so here it is ! I tried being as complete as I could, there are tests and docs with it !
Should I squash everything ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,6 +287,51 @@ slugArticle.getIdAttribute();
 // slug
 ```
 
+### `Schema.prototype.mappedBy(foreignKey)`
+
+This method permits to link a Schema property with a foreign property in another one Schema.
+
+```javascript
+const project = new Schema('projects');
+const user = new Schema('users');
+
+project.define({
+  collaborators: user.mappedBy('projects'),
+});
+
+user.define({
+  projects: user.mappedBy('collaborators'),
+});
+```
+
+This way, the relations will always be consistent. If you get this response for a user :
+
+```
+{
+  id: 1,
+  projects: [1, 2],
+}
+```
+
+Normalizr will give you this :
+
+```javascript
+{
+  result: [1],
+  entities: {
+    users: {
+      1: { id: 1, projects: [1, 2] }
+    },
+    projects: {
+      1: { id: 1, collaborators: [1] },
+      2: { id: 1, collaborators: [1] }
+    }
+  }
+}
+```
+
+Using this property is completely optional and you can even chose to use it only on one side of the relationship. // à vérifier
+
 ### `arrayOf(schema, [options])`
 
 Describes an array of the schema passed as argument.
--- a/README.md
+++ b/README.md
@@ -330,7 +330,7 @@ Normalizr will give you this :
 }
 ```
 
-Using this property is completely optional and you can even chose to use it only on one side of the relationship. // à vérifier
+Using this property is completely optional and you can even chose to use it only on one side of the relationship.
 
 ### `arrayOf(schema, [options])`
 
--- a/src/IterableSchema.js
+++ b/src/IterableSchema.js
@@ -18,4 +18,9 @@ export default class ArraySchema {
   getItemSchema() {
     return this._itemSchema;
   }
+
+  mappedBy(foreignKey) {
+    this._mappedBy = foreignKey;
+    return this;
+  }
 }
--- a/src/index.js
+++ b/src/index.js
@@ -15,18 +15,15 @@ function visitObject(obj, schema, bag, options) {
   let normalized = {};
   for (let key in obj) {
     if (obj.hasOwnProperty(key)) {
-      const entity = visit(obj[key], schema[key], bag, {
-        ...options,
-        parentId: schema.getId ? schema.getId(obj) : undefined,
-      });
+      const entity = visit(obj[key], schema[key], bag, options, typeof schema.getId === 'function' && schema.getId(obj));
       assignEntity.call(null, normalized, key, entity, obj);
     }
   }
   return normalized;
 }
 
-function defaultMapper(iterableSchema, itemSchema, bag, options) {
-  return (obj) => visit(obj, itemSchema, bag, options);
+function defaultMapper(iterableSchema, itemSchema, parentId, bag, options) {
+  return (obj) => visit(obj, itemSchema, bag, options, parentId);
 }
 
 function polymorphicMapper(iterableSchema, itemSchema, bag, options) {
@@ -37,12 +34,12 @@ function polymorphicMapper(iterableSchema, itemSchema, bag, options) {
   };
 }
 
-function visitIterable(obj, iterableSchema, bag, options) {
+function visitIterable(obj, iterableSchema, parentId, bag, options) {
   let itemSchema = iterableSchema.getItemSchema();
   if(iterableSchema._mappedBy) {
     itemSchema = clone(itemSchema).mappedBy(iterableSchema._mappedBy);
   }
-  const curriedItemMapper = defaultMapper(iterableSchema, itemSchema, bag, options);
+  const curriedItemMapper = defaultMapper(iterableSchema, itemSchema, parentId, bag, options);
 
   if (Array.isArray(obj)) {
     return obj.map(curriedItemMapper);
@@ -77,7 +74,7 @@ function defaultMergeIntoEntity(entityA, entityB, entityKey) {
   }
 }
 
-function visitEntity(entity, entitySchema, bag, options) {
+function visitEntity(entity, entitySchema, parentId, bag, options) {
   const { mergeIntoEntity = defaultMergeIntoEntity } = options;
 
   const entityKey = entitySchema.getKey();
@@ -97,23 +94,23 @@ function visitEntity(entity, entitySchema, bag, options) {
 
   if(entitySchema._mappedBy) {
     const fk = entitySchema._mappedBy;
-    if(entitySchema[fk] instanceof IterableSchema && options.parentId) {
+    if(entitySchema[fk] instanceof IterableSchema && parentId) {
       if(!stored[fk]) {
         stored[fk] = [];
       }
       stored[fk] = Array.from(new Set([
         ...stored[fk],
-        options.parentId
+        parentId
       ]));
-    } else if (options.parentId && !stored[fk]) {
-      stored[fk] = options.parentId;
+    } else if (parentId && !stored[fk]) {
+      stored[fk] = parentId;
     }
   }
 
   return id;
 }
 
-function visit(obj, schema, bag, options) {
+function visit(obj, schema, bag, options, parentId) {
   if (!isObject(schema)) {
     return obj;
   }
@@ -127,9 +124,9 @@ function visit(obj, schema, bag, options) {
   }
 
   if (schema instanceof EntitySchema) {
-    return visitEntity(obj, schema, bag, options);
+    return visitEntity(obj, schema, parentId, bag, options);
   } else if (schema instanceof IterableSchema) {
-    return visitIterable(obj, schema, bag, options);
+    return visitIterable(obj, schema, parentId, bag, options);
   } else if (schema instanceof UnionSchema) {
     return visitUnion(obj, schema, bag, options);
   } else {
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,7 @@
 import EntitySchema from './EntitySchema';
 import IterableSchema from './IterableSchema';
 import UnionSchema from './UnionSchema';
+import clone from 'lodash/clone';
 import isEqual from 'lodash/isEqual';
 import isObject from 'lodash/isObject';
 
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -287,6 +287,51 @@ slugArticle.getIdAttribute();
...
+    },
+    projects: {
+      1: { id: 1, collaborators: [1] },
+      2: { id: 1, collaborators: [1] }
+    }
+  }
+}
+```
+
+Using this property is completely optional and you can even chose to use it only on one side of the relationship. // à vérifier
```
RaitoBezarius: French ! :D

YoruNoHikage: Ooopsy, I'll remove it :P

unindented:
Yup, this looks great. Thank you @YoruNoHikage.

src/index.js
```diff
@@ -88,18 +92,41 @@ function visitEntity(entity, entitySchema, bag, options) {
   let normalized = visitObject(entity, entitySchema, bag, options);
   mergeIntoEntity(stored, normalized, entityKey);
 
+  if(entitySchema._mappedBy) {
+    const fk = entitySchema._mappedBy;
+    if(entitySchema[fk] instanceof IterableSchema && parentId) {
+      if(!stored[fk]) {
+        stored[fk] = [];
+      }
+      stored[fk] = Array.from(new Set([
```
gaearon: Why are we using ES6 features here? As far as I know this project used to work fine in older environments, but it won’t anymore.

src/IterableSchema.js
```diff
@@ -18,4 +18,9 @@ export default class ArraySchema {
   getItemSchema() {
     return this._itemSchema;
   }
+
+  mappedBy(foreignKey) {
+    this._mappedBy = foreignKey;
```
gaearon: It seems a bit inconsistent that the function uses a fluent API (as in, returns this while mutating the original object) while there is not a single such precedent in the existing API. What does `schema.mappedBy('a').mappedBy('b')` mean? Fluent APIs are confusing when they are mutating, and I think we should have cloned the schema here instead.

src/index.js
YoruNoHikage: Oh right, I used this to avoid duplicates, surely there is a better alternative.

gaearon: It’s good to refer to MDN for such cases, e.g. https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Set

screen shot 2016-04-07 at 16 13 17

gaearon: Same for https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from:

screen shot 2016-04-07 at 16 14 13

src/index.js
```diff
@@ -33,9 +34,12 @@ function polymorphicMapper(iterableSchema, itemSchema, bag, options) {
   };
 }
 
-function visitIterable(obj, iterableSchema, bag, options) {
-  const itemSchema = iterableSchema.getItemSchema();
-  const curriedItemMapper = defaultMapper(iterableSchema, itemSchema, bag, options);
+function visitIterable(obj, iterableSchema, parentId, bag, options) {
+  let itemSchema = iterableSchema.getItemSchema();
+  if(iterableSchema._mappedBy) {
```
gaearon: Nit: existing code puts spaces after `if`

src/index.js
```diff
@@ -88,18 +92,41 @@ function visitEntity(entity, entitySchema, bag, options) {
   let normalized = visitObject(entity, entitySchema, bag, options);
   mergeIntoEntity(stored, normalized, entityKey);
 
+  if(entitySchema._mappedBy) {
+    const fk = entitySchema._mappedBy;
+    if(entitySchema[fk] instanceof IterableSchema && parentId) {
```
gaearon: Nit: existing code puts spaces after `if`

src/index.js
```diff
@@ -33,9 +34,12 @@ function polymorphicMapper(iterableSchema, itemSchema, bag, options) {
   };
 }
 
-function visitIterable(obj, iterableSchema, bag, options) {
-  const itemSchema = iterableSchema.getItemSchema();
-  const curriedItemMapper = defaultMapper(iterableSchema, itemSchema, bag, options);
+function visitIterable(obj, iterableSchema, parentId, bag, options) {
+  let itemSchema = iterableSchema.getItemSchema();
+  if(iterableSchema._mappedBy) {
+    itemSchema = clone(itemSchema).mappedBy(iterableSchema._mappedBy);
```
gaearon: Do we need a *deep* clone here? Does this library need to depend on a very complex `clone()` implementation for the sake of this line? Wouldn’t a shallow clone be enough here?

It seems like if we don’t make `mappedBy` mutating in the first place, we can drop dependency on `clone()`.

unindented:
Wow, I did a shit job reviewing this PR. 😭

src/IterableSchema.js
YoruNoHikage: Something like this could work ?
```js
return {
  Object.create(this),
  this,
  { _mappedBy: foreignKey }
};
```

gaearon:
I reverted for now to avoid confusion (#88). This needs a little bit more work :smile: 
No worries @unindented, I don’t really have much time for this repo so I appreciate you maintaining with it. There’s just a few concerns I have:

* We shouldn’t arbitrarily break older browsers for the sake of adding new features.
* We should try to prevent the API from becoming a swiss knife for “everything anyone ever wanted” that is impossible to understand or modify due to all the small edge cases.

I appreciate that people want this feature but I would like to see a little bit more effort into making it consistent with the current API (e.g. no fluent methods), as well as more thought put into:

* How will people want to extend this later?
* Are we just picking a particular simple use case? Are we going to be bombarded with requests for more complex variations of this once it becomes part of the API?
* Do we really want to expand the scope of what the library does, and how can we do this in the most minimal way so we don’t screw up with this feature later down the line?
* Does this conflict with existing features? (e.g. now we seem to have two ways of defining relationships—via `define()` and via back reference. which should be used when? it’s not clear.)

Libraries get bloated quickly and eventually nobody understands how they work. I’d be sad to see this library evolve into a kitchen sink ORM that causes people to pull out their hair when there is a bug but the code is impossible to walk through.

Ultimately I defer to you to merge and decide whether this feature is worth the complexity it introduces. I’m happy to see you merge another PR that fixes browser compatibility issues, but I’d prefer to see more API discussions of pros and cons before a variant of this gets merged. Thanks!

unindented:
I agree. I felt the pressure of having to merge this because of all the positive comments, and didn't judge it as critically as I should have. My lack of experience maintaining medium-sized OSS projects shows.

I'll let @YoruNoHikage comment on possible alternatives that align better with the existing APIs, but I agree with not merging this in its current form.

I'll revert https://github.com/gaearon/normalizr/pull/56 and https://github.com/gaearon/normalizr/commit/81d95ddedaba7d0eb8e71792d3f5c91a7ca25a44 too, as they add a fluent API that doesn't help.

gaearon:
The problem with fluent APIs is they don’t make it clear whether they are mutating or not. `change(): void` is obviously mutating. `change(): T` in an immutable-friendly community (which React/Redux community is) usually signals that `change()` returns a copy. `change(): T` that just returns `this` and mutates the object seems like the worst of both worlds because it trades a few keystrokes for hard-to-find bugs.

src/index.js
downsider: Correct me if I'm wrong, but normalizr uses `class` and several other ES6 features already. These are compiled down to ES5 using babel, so there shouldn't be a problem with `Array.from()` and `new Set()`, surely?

YoruNoHikage: In any case, it must be too much for what I want to achieve. This could work too : 
```js
stored[fk] = stored[fk] || [];
if (stored[fk].indexOf(parentId) === -1) {
  stored[fk] = [...stored[fk], parentId];
}
```

YoruNoHikage:
I've commented possible solutions, I can integrate them and refactor a bit. I may be too biased but this feature is pretty simple not enforcing anyone to do anything. :smile:

gaearon:
Considering the use case in https://github.com/gaearon/normalizr/issues/70:

>Now you want to create a project, you send some request to a server that replies with the project entity like this:
>```js
>{
>  id: 80,
>  contributors: [1], // your userId for example
>  // other fields
>}
>```
>You normalize the response and merge with the rest of your entities (just like the real-world example for Redux), the projects' entities are updated but not the user.

I think that naïvely keeping them in sync might work for simple cases but will cause significant problems later for any app that is serious about handling its data, and making it too easy to do the wrong thing would hurt normalizr users in the long term.

It is easier to see this as a real problem when you consider what happens when the data set becomes larger. What if your app usage grows, and `user.projects` and `project.contributors` become paginated?

There are probably different parts of UI where you show lists of either. **You’ll want to paginate them separately because, while you may want to load first 20 contributors of some projects, you may also want to independently load first 20 projects of some users.** At this point it becomes obvious that it’s not just about keeping some arrays in sync: you want to keep completely independent state for each paginated collection, and this is outside of normalizr’s scope, at least the way I see it.

In the project I worked on (which was the reason I wrote normalizr), we kept *entities* normalized in the state. Any changes to entities would get merged directly with a naïve merge algorithm. The state of the paginated collections lived elsewhere in the state tree, and while the relevant responses were also processed by normalizr to “extract” the entities, the logic of adding or removing items from collections was managed by reducers [very much like this one](https://github.com/reactjs/redux/blob/master/examples/real-world/reducers/paginate.js), and was mostly governed by actions (both optimistic and response actions).

Sure, sometimes we used the simple scheme that this PR attempts to support better. We’d put arrays of IDs directly as entity fields when we *knew* they wouldn’t get larger, or didn’t care enough. However in this case we would make sure that server responses contained _all_ the necessary info to update these entities. **We would not allow the server to return just IDs without corresponding entities.**

In other words, adding a user to a project would return both the `user`, the `project`, and any related users and projects that happen to be referenced in the arrays, thus ensuring all entities are updated correctly.

Does this sound like a large response? We only used this technique when we could get away with it. When we noticed the responses grow large, we would begin to omit `contributors` (or user’s `projects`) from the response completely, and instead add some code to the reducers to update `user.projects` and `project.contributors` in response to specific actions (e.g. `ADD_PROJECT_REQUEST` and `ADD_PROJECT_SUCCESS`). Note that you’ll want to do this anyway if you care about showing optimistic updates!

Finally, we would eventually switch to separate collections going in both directions elsewhere in the state tree, that get populated by different paginated requests. Again, they can tune to the same actions and update the pagination state optimistically if they care to.

I’m not saying our approach was the best one, but I thought it might be worthwhile to provide some context on why I didn’t include such seemingly “basic” feature out of the box. Maybe it’s worth including but I can see how adding support for this will increase the scope of the library and potentially turn it from a “library that’s good at normalizing responses” to a “ORM that’s bad at handling paginated data” just because we make a first-class API for patterns that aren’t very solid.

RaitoBezarius:
Thanks @gaearon for your insights, it's really interesting, because I'm encountering this problem which `mappedBy` attempted to solve.

It's true that managing paginated states is going to be hard if we do this. But, could I ask how is managed your denormalized state then?

> Finally, we would eventually switch to separate collections going in both directions elsewhere in the state tree, that get populated by different paginated requests

If you have separate collections, do you have denormalized state?
Sorry if I misunderstood you, but I'm really interested in how to achieve proper denormalization while tracking correctly the relations.

A project which implemented a more or less similar solution is: https://github.com/itsmepetrov/redux-entities/

Do you think that another library on the top of `normalizr` should be required? Funny, we have an "ORM" in Redux: https://github.com/tommikaikkonen/redux-orm

Thanks for your insights!

gaearon:
@RaitoBezarius For paginated collections, my state looks like

```js
{
  entities: {
    users: {
      1: { id: 1, name: 'Dan' },
      2: { id: 2, name: 'Andrew' },
      ...
    },
    repos: {
      1: { id: 1, name: 'Redux' },
      ...
    },
  },
  userRepos: {
    1: {
      isFetching: true,
      ids: [],
      nextPageUrl: null,
    },
    2: {
      isFetching: false,
      ids: [1, 2, 3, 4],
      nextPageUrl: '/users/2/repos?page=2',
    },
    ...
  },
  repoContributors: {
    1: {
      isFetching: false,
      ids: [1, 2],
      nextPageUrl: null,
    },
    2: {
      isFetching: true,
      ids: [],
      nextPageUrl: '/repos/2/contributors?page=2',
    },
    ...
  }
}
```

So they still use normalized IDs from entities, they are just not stored directly inside those entities.

kristojorg:
@gaearon First off let me say that I appreciate everything you've done for redux, normalizr and react. I have learned so much from all of your work here. Immensely helpful. 

I'm currently bringing Redux and React to our web app and I would love some advice. We have many moving pieces, and I'm struggling to understand the best way to design our state tree (and then to normalize our API response into that tree. So:

1. When do I break these out into entities, and what do I use collections for (as you have above) We have:
  - Users, who can be coaches or trainees of other users
  - Groups which have users
  - Posts which are in a user stream or group stream and authored by a user
  - Comments which are on a post
  - Calendar activities which are on a user or group and similarly owned by a user

2. How do I normalize a JSON response where the id of an entity is nested inside the object itself:
`javascript
{
  myID: 1,
  myTrainees:{
    id: 1 //id of their trainee-expert relation
    traineeUser:{
      id:1 //id of the actual trainee which should be recorded as a user
    }
  }
}
`
I suppose I can do it manually when handling the response by looping over them with lodash but I'm curious if this is a case supported by normalizr. 

Thank you !!

gaearon:
This would better belong on StackOverflow than in an issue discussion. Thanks!

PR #82, opened 2016-03-30
Fix Safari GC issue by removing mapValues() in favor of Object.keys().reduce()

paularmstrong:
# Problem

See the gist: [Safari GC Issue: Normalizr + Lodash](https://gist.github.com/paularmstrong/02330d28dbe202a3aa8cd688473a1919) (credit to @alunny)

Summary: normalizr + lodash in Safari appears to encounter a problem sometimes in which Safari's garbage collector interrupts the entity mapping, causing the entities object to be empty.

# Solution

This is sort of a half-way-there solution, because the ultimate goal would be to remove all production dependencies. However, `isEqual`/deep-equality is not an easy problem to solve and lodash's implementation is the best available.

* Removes `lodash` dependency
* Uses `lodash.isequal` package for `isEqual`
* Adds an `isObject` util method
* Refactors use of `lodash/mapValues`

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,0 +1,4 @@
+export function isObject(value) {
+  var type = typeof value;
+  return !!value && (type === 'object' || type === 'function');
+}
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,4 +1,0 @@
-export function isObject(value) {
-  var type = typeof value;
-  return !!value && (type === 'object' || type === 'function');
-}
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "babel-core": "^5.8.3",
     "babel-loader": "^5.3.2",
     "chai": "^3.2.0",
+    "lodash": "^4.6.1",
     "mocha": "^2.2.5",
     "rimraf": "^2.4.2",
     "webpack": "^1.10.5"
--- a/package.json
+++ b/package.json
@@ -36,12 +36,11 @@
     "babel-core": "^5.8.3",
     "babel-loader": "^5.3.2",
     "chai": "^3.2.0",
-    "lodash": "^4.6.1",
     "mocha": "^2.2.5",
     "rimraf": "^2.4.2",
     "webpack": "^1.10.5"
   },
   "dependencies": {
-    "lodash.isequal": "^4.1.1"
+    "lodash": "^4.0.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -41,6 +41,6 @@
     "webpack": "^1.10.5"
   },
   "dependencies": {
-    "lodash": "^4.0.0"
+    "lodash.isequal": "^4.1.1"
   }
 }
--- a/src/IterableSchema.js
+++ b/src/IterableSchema.js
@@ -1,4 +1,4 @@
-import isObject from 'lodash/isObject';
+import { isObject } from './utils';
 import UnionSchema from './UnionSchema';
 
 export default class ArraySchema {
--- a/src/IterableSchema.js
+++ b/src/IterableSchema.js
@@ -1,4 +1,4 @@
-import { isObject } from './utils';
+import isObject from 'lodash/isObject';
 import UnionSchema from './UnionSchema';
 
 export default class ArraySchema {
--- a/src/UnionSchema.js
+++ b/src/UnionSchema.js
@@ -1,4 +1,4 @@
-import isObject from 'lodash/isObject';
+import { isObject } from './utils';
 
 export default class UnionSchema {
   constructor(itemSchema, options) {
--- a/src/UnionSchema.js
+++ b/src/UnionSchema.js
@@ -1,4 +1,4 @@
-import { isObject } from './utils';
+import isObject from 'lodash/isObject';
 
 export default class UnionSchema {
   constructor(itemSchema, options) {
--- a/src/index.js
+++ b/src/index.js
@@ -1,8 +1,8 @@
 import EntitySchema from './EntitySchema';
 import IterableSchema from './IterableSchema';
 import UnionSchema from './UnionSchema';
-import isEqual from 'lodash.isequal';
-import { isObject } from './utils';
+import isEqual from 'lodash/isEqual';
+import isObject from 'lodash/isObject';
 
 function defaultAssignEntity(normalized, key, entity) {
   normalized[key] = entity;
--- a/src/index.js
+++ b/src/index.js
@@ -1,9 +1,8 @@
 import EntitySchema from './EntitySchema';
 import IterableSchema from './IterableSchema';
 import UnionSchema from './UnionSchema';
-import isObject from 'lodash/isObject';
-import isEqual from 'lodash/isEqual';
-import mapValues from 'lodash/mapValues';
+import isEqual from 'lodash.isequal';
+import { isObject } from './utils';
 
 function defaultAssignEntity(normalized, key, entity) {
   normalized[key] = entity;
@@ -41,7 +40,10 @@ function visitIterable(obj, iterableSchema, bag, options) {
   if (Array.isArray(obj)) {
     return obj.map(curriedItemMapper);
   } else {
-    return mapValues(obj, curriedItemMapper);
+    return Object.keys(obj).reduce(function (objMap, key) {
+      objMap[key] = curriedItemMapper(obj[key]);
+      return objMap;
+    }, {});
   }
 }
 


src/utils.js
```diff
@@ -0,0 +1,4 @@
+export function isObject(value) {
+  var type = typeof value;
+  return !!value && (type === 'object' || type === 'function');
```
gaearon: This is more naïve than Lodash impl. For example this will return `true` for arrays.

gaearon:
>because the ultimate goal would be to remove all production dependencies.

Why is this a goal?

gaearon:
cc @jdalton.
IMO if there is an issue with Lodash, we should file it there rather than try to work around it here.

paularmstrong:
> > because the ultimate goal would be to remove all production dependencies.

>Why is this a goal?

Needing to pull in an extra huge dependency for such a small library doesn't seem totally necessary.

As for filing at lodash, I'm not totally certain there's a real problem there either, and we haven't found a repro using just lodash. It's also localized to Safari only, so the source of the problem is in Safari/Webkit's GC. We're working on filing a radar for the issue.

gaearon:
>Needing to pull in an extra huge dependency for such a small library doesn't seem totally necessary.

What is huge? If we need a deep equality check, presumably implementing it in house would be no smaller than Lodash’s implementation. We don’t pick any other parts of Lodash we don’t need—it’s not like we depend on the library itself. We depend on individual modules, only they will get bundled by the app.

paularmstrong:
We could keep lodash if you really want, but I'm still not convinced on using `mapValues`, as the replacement solution is short and sweet.

Also, when I built from the repro gist provided, all of lodash is pulled in from normalizr, not just the methods needed.

gaearon:
>Also, when I built from the repro gist provided, all of lodash is pulled in from normalizr, not just the methods needed.

I think it’s not all of Lodash—it’s all the modules required by `isEqual` impl. It spans multiple modules but it’s still far from all of Lodash.

src/utils.js
jdalton: Naw that's straight up Lodash's implementation. It's [straight copypasta](https://github.com/lodash/lodash/blob/4.6.1-npm/isObject.js).

paularmstrong:
Alright, I can revert to use lodash for `isObject`. 

Still concerned about mapValues, though?

gaearon:
Yea I don’t have a strong opinion on `mapValues`. Fine with using `reduce` there. But we shouldn’t remove Lodash just for the sake of removing Lodash. We cherry-pick only the modules we need and I don’t see the point of maintaining small-but-easy-to-get-it-wrong modules like `isObject`.

src/utils.js
gaearon: Oh, my bad. Thanks for correcting!

jdalton:
I don't get it. Lodash offers utils, you all repeatedly rewrite them. OK.

jdalton:
Is it a like project culture thing? I don't get it.

gaearon:
I don’t want to rewrite anything or maintain it :smile: 

@paularmstrong Can you help me understand why pulling out Lodash and using the same implementation solves the issue in Safari?

gaearon:
It appears like the only actual change here would be replacing `mapValues` with `reduce`. Is this what solves the Safari issue? Otherwise we're still using Lodash’s `isEqual`, and the `isObject` implementation you added seems to match Lodash’s.

paularmstrong:
@gaearon yes. Replacing mapValues solves the issue. It's very hard to pinpoint whether the issue is in lodash or in the combination of normalizer and lodash at the moment.

This patch is in the interest of fixing what may be a pretty prevalent issue for those using normalizr first, since it's where the issue is cropping up.

alunny:
Here is the repro in a jsbin: https://output.jsbin.com/coquxaduhu

If you open in Safari, without inspector open (which frustratingly appears to affect the behavior), you will see "missing entities.tweets" logged after some time. It's nondeterministic. We haven't been able to repro in any other branch.

We're working on isolating the error and reporting to Apple, but until then, empirically, @paularmstrong's patch causing the error to stop happening.

gaearon:
Btw I appreciate both @paularmstrong’s help maintaining this repo and @jdalton’s help maintaining all the utilities I never want to write again :smile: . I’m sure we’ll find a solution but if it affects Lodash’s `mapValues` implementation we should *really* find a way to fix it upstream because it affects tons of other packages too in this case.

jdalton:
> (which frustratingly appears to affect the behavior)

That's usually an indicator of a JIT bug.

The `mapValues` method is super light compared to the work done by `isEqual` which recursively crawls objects and tracks visited ones (lots more things for a GC to get grump about)

gaearon:
Confirmed I can reproduce this with Safari in JSBin posted above.

jdalton:
This is less likely a Lodash issue and more likely using built-in reduce disqualifies it from some JIT optimizations so bypasses the bug.

jdalton:
Previous versions of Safari had a JIT issue with `.length` which effected jQuery, Underscore, Lodash.
@stefanpenner did you hit something similar?

gaearon:
I’d like to keep this open for a few more hours if you’d like to investigate more. I think we should release a fix in the evening since it consistently helps at least a couple of people but it’s not really ideal.

jdalton:
Is there a way to get a reduced repro (without all the bundled stuff). Usually these things can come down to doing something in a tight loop one way then after the loop providing different input and it returning the wrong result.

For example:
```js
    QUnit.test('should iterate over an object with numeric keys (test in Mobile Safari 8)', function(assert) {
      assert.expect(1);

      // Trigger a Mobile Safari 8 JIT bug.
      // See https://github.com/lodash/lodash/issues/799.
      var counter = 0,
          object = { '1': 'foo', '8': 'bar', '50': 'baz' };

      lodashStable.times(1000, function(assert) {
        _.filter([], alwaysTrue);
      });

      _.filter(object, function() {
        counter++;
        return true;
      });

      assert.strictEqual(counter, 3);
    });

    QUnit.test('should work with host objects in IE 8 document mode (test in IE 11)', function(assert) {
      assert.expect(2);

      // Trigger a Chakra JIT bug.
      // See https://github.com/jashkenas/underscore/issues/1621.
      lodashStable.each([body, xml], function(object) {
        if (object) {
          lodashStable.times(100, _.isFunction);
          assert.strictEqual(_.isFunction(object), false);
        }
        else {
          skipAssert(assert);
        }
      });
    });
```

jdalton:
If I had to guess I'd suspect the [_createBaseFor](https://github.com/lodash/lodash/blob/4.6.1-npm/_createBaseFor.js) as being a possible place for the JIT bug to occur.

jdalton:
It would help to know how you're calling `mapValues` (what kind of input params).

paularmstrong:
> It would help to know how you're calling mapValues (what kind of input params).

Plain JSON object, curried function wrapper

jdalton:
Does the object have a `length` property or numeric keys?
Could you provide a small sample of the object and the curried function.

jdalton:
\cc @dmethvin @megawac New Safari JIT bug.

necolas:
Friendly reminder that we're using this in production on mobile.twitter.com and it's causing production issues for Safari users. This patch to normalizr is trivial and a simplification with no apparent loss of functionality.

> I think it’s not all of Lodash

Paul is pointing to the fact that when you [install normalizr](https://github.com/gaearon/normalizr/blob/47b2873113da994d03e61a002451f0ca1a4c54b8/package.json#L44), it installs the whole of lodash.

jdalton:
@necolas The JIT issue has been around for a few weeks (or that's how long I've heard grumbles about it) though no ones produced a simplified repro. Lodash is prepping for a version bump tonight so if you help create a simplified repro it'd be fixed in hours.

dmethvin:
Apple just announced the [Safari Technology Preview](http://www.anandtech.com/show/10199/apple-announces-the-safari-technology-preview), maybe that will lift the curtain a bit and we can at least see if their canary-like build fixes this.

jdalton:
\cc WebKit folks @rniwa @grorg. Any known issues?

alunny:
I have a smaller (still 130K) repro that may of use to the webkit folk:
https://gist.github.com/alunny/5a42d0e667fbf5fc498566db4d9a7ea5

That removes all of the async code, so you can run it with `jsc` and repro the issue (and also run with `node` to see the expected behavior).

I'll be out most of this afternoon, but will be back in the evening to hammer on it some more.

jdalton:
More context for WebKit folks:
Lodash's `mapValues` will use `Object.keys` under the hood.

gaearon:
>Paul is pointing to the fact that when you install normalizr, it installs the whole of lodash.

Yeah but that’s just how npm works. Aside from a bigger download in `node_modules`, I don’t really see any issues with this.

gaearon:
I’ll cut a patch release with this to help reduce the number of affected users.

jdalton:
I guess I'll punt too then. This will likely not be revisited after the merge.

rniwa:
Could someone describe to me what the symptom here is?  Is the issue that values in Object.keys don't return all keys?  Can I have a pointer to the actual implementation of mapValues?  I can't find it anywhere.

gaearon:
@jdalton Yeah in this case we likely won’t be revisiting. Thanks for your help!

jdalton:
@rniwa [mapValues](https://github.com/lodash/lodash/blob/4.6.1-npm/mapValues.js), [_baseForOwn](https://github.com/lodash/lodash/blob/4.6.1-npm/_baseForOwn.js), [_createBaseFor](https://github.com/lodash/lodash/blob/4.6.1-npm/_createBaseFor.js), and [keys](https://github.com/lodash/lodash/blob/4.6.1-npm/keys.js), & [_baseKeys](https://github.com/lodash/lodash/blob/4.6.1-npm/_baseKeys.js).

gaearon:
The interim fix is out in 2.0.1. Based on reports in this thread it is enough to work around the issue although obviously can’t give any guarantees.

cc @necolas

necolas:
Thanks for the quick turn around

> Yeah but that’s just how npm works.

Yep :)

rniwa:
@jdalton : Thanks!

megawac:
Yeah but that’s just how npm works. Aside from a bigger download in node_modules, I don’t really see any issues with this.

You can consider using the [lodash-modularized npm suite](https://www.npmjs.com/browse/keyword/lodash-modularized) to avoid installing all of lodash

gaearon:
Sorry, I just don’t see why bother. The result is identical both ways. User is more likely to already have `lodash` package than any of the modularized ones.

Anyway, if you’d like to discuss this, I welcome you to open a new issue. Let’s keep this thread focused on the Safari GC bug. Thanks!

jdalton:
@rniwa I modified [the repro](https://output.jsbin.com/coquxaduhu) to remove the `getIteratee` call from [mapValues](https://github.com/lodash/lodash/blob/4.6.1-npm/mapValues.js) and that seems to fix it.

jdalton:
The `baseIteratee` module is [here](https://github.com/lodash/lodash/blob/4.6.1-npm/_baseIteratee.js). Could it be related to the type juggling?

gaearon:
Another instance of the same issue: https://github.com/gaearon/react-proxy/issues/55

gaearon:
> the Safari JIT (I'm running Safari 9.0.3 (build 11601.4.4)) makes baseIteratee (called from lodash.find) randomly think that the find predicate ([key]) => key === Component -- is an Object and not a Function.

jdalton:
Changing the `typeof` check in [baseIteratee](https://github.com/lodash/lodash/blob/4.6.1-npm/_baseIteratee.js) to a heavier [isFunction](https://github.com/lodash/lodash/blob/4.6.1-npm/isFunction.js) check avoid the JIT bug. I don't want to do that though so I'll poke around a bit more but that looks like the issue.

paularmstrong:
Thanks for the help and extra digging, everyone!

jdalton:
@rniwa  The quick fix for me is to just use two `typeof `checks instead of storing the result in a variable. However there are several other places in Lodash that still store results in a variable so when the issue is narrowed down if there's a better workaround please let me know.

rniwa:
Thanks a lot everyone for the bug report!  With the information provided on this issue and https://github.com/gaearon/react-proxy/issues/55, I've determined that the bug is caused by our DFG JIT and  filed a [WebKit bug 156034](https://webkit.org/b/156034).  Subsequently, my colleague was able to pinpoint the culprit to a [single line of code](https://trac.webkit.org/browser/trunk/Source/JavaScriptCore/dfg/DFGAbstractInterpreterInlines.h?rev=198865#L1173) and he's working on a fix.

jdalton:
@rniwa Sweet! Do you know of a reduced repo for this? Does it require a check for "object"?

jdalton:
*repro.

rniwa:
Here's a minimum reproduction for those who are interested: https://gist.github.com/anonymous/49715c82e834f8e2c184d35a821ec468

Yes, this is a bug in `typeof`.  It can return "object" for a `Function`.  I *think* you can use `x instanceof Function` since the latter wouldn't suffer from the same bug.

alunny:
@rniwa thanks for following up so quickly, very much appreciated :)

jdalton:
Here's a simpler repro without juggling the input type:
https://output.jsbin.com/woniku/1

rniwa:
By the way, this bug is caused by our optimizing compiler generating wrong machine code, and nothing to do with GC on the contrary to what the title says.

jdalton:
This fix made it into Lodash [v4.7.0](https://github.com/lodash/lodash/wiki/Changelog#v470) (due out in the morning as you all don't want me publishing before I go to sleep)

jdalton:
BTW this could have also been avoided by using webpack to reroute `baseIteratee` module requests to `identity`. It would have also produced a smaller build without the need to reimplement methods.

justsml:
Thx @jdalton @rniwa et al.. I didn't believe the ghost stories about this bug!

jwpari:
Thanks everyone! This was a crazy heisenbug.

PR #84, opened 2016-04-06
Add type declarations

unindented:


unindented:
@blakeembrey could you please review?

blakeembrey:
@unindented Just remember that TypeScript is going to do structural type checking, so some of those values already appear to be assignable to each other.

typings.d.ts
```diff
@@ -0,0 +1,67 @@
+interface EntitySchemaOptions {
```
blakeembrey: You will probably want to export these types.

unindented: All types that are used by public APIs should be exported? So I should export everything in this file, right?

unindented:
@blakeembrey Forgive my ignorance, I'm very new to type systems. Could you give me an example to illustrate the fact that "some of those values already appear to be assignable to each other", and what consequences it has?

unindented:
@gaearon What are your thoughts on adding type definitions to the repo? I have an attempt at TypeScript definitions here, and could work on a `.flowconfig` file...

blakeembrey:
@unindented This is the basic gist: http://www.typescriptlang.org/play/index.html#src=class%20Foo%20%7B%7D%0A%0Aclass%20Bar%20%7B%7D%0A%0Afunction%20x%20(foo%3A%20Foo)%20%7B%7D%0A%0Ax(new%20Bar())%0Ax('test')%0Ax(10)%0Ax(Foo)

Since all the interfaces are structurally the same, you aren't actually doing any type checking.

typings.d.ts
blakeembrey: Pretty much, export everything.

unindented:
@blakeembrey Aha, understood. Adding empty interfaces is almost equivalent to `any`, so my union type was useless.

I've updated the PR with proper types.

unindented:
@paularmstrong @gaearon do you think the type definitions belong to this repo? There are ways to host the type definitions in a separate repo, but then we wouldn't be able to reference them in the `package.json`, and it would add some friction for users.

paularmstrong:
I don't see why normalizr shouldn't include its own type declarations. However, is there a way to have an automated test ensure that these are updated when necessary?

unindented:
Umm, I don't know the best way to ensure type declarations are up-to-date. @blakeembrey what do you suggest?

gaearon:
>However, is there a way to have an automated test ensure that these are updated when necessary?

If we add them, we should add tests for them. 
I’m also ignorant of the process, here’s what community did with Redux:

https://github.com/reactjs/redux/pull/1413
https://github.com/reactjs/redux/pull/1526
https://github.com/reactjs/redux/pull/1537
https://github.com/reactjs/redux/pull/1545

These include tests.

unindented:
Ok, I'll follow the same approach.

blakeembrey:
Yes, that's what we usually do too. You can have some valid code and make sure it compiles validly with `tsc`, or even run some of the test suite with TypeScript.

unindented:
Hey @paularmstrong, I finally got around to this.

I've updated the definitions with the latest changes that were made (`getMeta`, etc.) and added tests that cover all signatures.

Let me know what you think!

paularmstrong:
Thanks @unindented. I'll merge this soon for v2.2.0

unindented:
@paularmstrong any planned date for `v2.2.0`?

paularmstrong:
Been on vacation. Maybe can shoot for late next week. If there's a strong request, though, I could defer the docs/examples/tests updates that we need.

unindented:
Tests are already in this PR. Other projects, like `redux`, assume people who are using TypeScript know how to consume types, so they don't add specific docs or examples.

paularmstrong:
@unindented I mean #105, #106, and #107

PR #89, opened 2016-04-07
provide schema in assignEntity

frankwallis:
This PR is a small change to provide ```schema``` as an argument to the ```assignEntity``` function. 

My use-case is that when normalizing entities I want the key to be present under a different name in certain circumstances. I need to support both normalized and denormalized data and having fields with the same name but different types is causing quite a lot of confusion and noise.

With this change I can flag the schemas which I want to be assigned with a new name and then override the assignEntity method. See the added test for an example of this working.

Happy to discuss or make changes if you want, many thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,9 @@ const article = new Schema('articles', { idAttribute: 'slug' });
 // Or you can specify a function to infer it
 function generateSlug(entity) { /* ... */ }
 const article = new Schema('articles', { idAttribute: generateSlug });
+
+// You can also specify meta properties to be used for customizing the output in assignEntity (see below)
+const article = new Schema('articles', { idAttribute: 'slug', meta: { removeProps: ['publisher'] }});
 ```
 
 ### `Schema.prototype.define(nestedSchema)`
--- a/src/EntitySchema.js
+++ b/src/EntitySchema.js
@@ -9,6 +9,7 @@ export default class EntitySchema {
     const idAttribute = options.idAttribute || 'id';
     this._getId = typeof idAttribute === 'function' ? idAttribute : x => x[idAttribute];
     this._idAttribute = idAttribute;
+    this._meta = options.meta;
   }
 
   getKey() {
@@ -21,6 +22,13 @@ export default class EntitySchema {
 
   getIdAttribute() {
     return this._idAttribute;
+  }
+
+  getMeta(prop) {
+    if (!prop || typeof prop !== 'string') {
+      throw new Error('A string non-empty property name is required');
+    }
+    return this._meta && this._meta[prop];
   }
 
   define(nestedSchema) {
--- a/src/index.js
+++ b/src/index.js
@@ -15,7 +15,7 @@ function visitObject(obj, schema, bag, options) {
   for (let key in obj) {
     if (obj.hasOwnProperty(key)) {
       const entity = visit(obj[key], schema[key], bag, options);
-      assignEntity.call(null, normalized, key, entity, obj);
+      assignEntity.call(null, normalized, key, entity, obj, schema);
     }
   }
   return normalized;
--- a/test/index.js
+++ b/test/index.js
@@ -108,7 +108,7 @@ describe('normalizr', function () {
     Object.freeze(input);
 
     var options = {
-      assignEntity: function(obj, key, val) {
+      assignEntity: function(obj, key, val, originalInput, schema) {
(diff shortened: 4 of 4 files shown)


frankwallis:
I'm keen to know if this is likely to be merged, I read with interest [this comment](https://github.com/gaearon/normalizr/pull/80#issuecomment-206957386) about not wanting normalizr to become a 'bloated kitchen sink ORM' and I completely agree with those sentiments. I believe that the solution to that problem may be to provide richer extension points (```assignEntities``` and ```mergeEntities``` being the existing ones), specifically having access to the schema in those methods would allow users to add custom fields onto the schema which they can use to control the output. This means that the configuration is still held centrally within the schema structure rather than being split into a separate processing pipeline.

I  hope this makes sense, and if you have any comments about the PR I look forward to hearing them.

paularmstrong:
I'm concerned that this is a very localized issue. It's also possible that I don't understand the problem. However, your test case doesn't really show why I'd want this either.

frankwallis:
I agree that this is a very localized issue, but the change does not just support this specific issue, instead it allows all users to receive the schema in the ```assignEntities``` extension point. They will then be able to customise the output from normalizr more easily by adding extra properties/flags onto the schema objects.

In my specific case I don't want a situation where sometimes ```client.address``` contains a nested object and sometimes it contains an address id (for example). Having one field contain different types at different times in the application can be confusing.

To get around this when normalizing with the ```address``` schema I can specify on the schema a new field-name for the normalized id. In the test example I provide a prefix so that the id is normalized to ```client.__address``` but I could equally well provide a new field name so that it comes out as ```client.addressId``` or set a flag so that it is parsed as an integer or whatever I want to happen.

frankwallis:
Is there anything more I can do to get this merged? If the test I added is not a clear example of how this extra parameter can be useful then I will happily update it. 

I essentially just mirrored the existing test for passing the whole object as an extra parameter to ```assignEntities``` as my change is very similar, but if it is not clear then I will rework it.

paularmstrong:
I'm still on the fence (a bit in the negative side) for merging this. I'd like to see signoff from @unindented and/or @gaearon.

unindented:
I'm with @paularmstrong. This doesn't look like a normal use case, so I'd suggest you solve it by pre-processing the payload, or some other means external to `normalizr`.

gaearon:
>This doesn't look like a normal use case, so I'd suggest you solve it by pre-processing the payload

I agree.

frankwallis:
I accept that my use case is probably unique to me, but is there a specific reason you do not want to provide the schema as an argument to ```assignEntities```? You currently provide the whole original object as a parameter, presumably for some other "abnormal" use-case, I don't really see what the difference is?

gaearon:
I see writing and reading arbitrary fields to and from schema potentially problematic. What if we add a `prefix` field to it as a part of some refactoring? This would break your code. If this was implemented, it would need to be more resilient to changes to internal implementation.

frankwallis:
Yes I was thinking about having a dedicated field on the schema, where users could put their custom properties, maybe called "custom", would you be open to that?

gaearon:
`new Schema(key, [options])` already accepts an `options` argument so maybe we can add a `meta` field to it.

frankwallis:
Sounds good to me, I will make the changes and resubmit, thank you for listening!

frankwallis:
I have updated the PR so that the ```options``` parameter to the constructor of ```EntitySchema``` can have a ```meta``` property to hold user defined properties. I have also updated the test to demonstrate how it can be used to remove specific properties from the normalized output.

I am not quite sure how much you want me to add to the docs but I have added a line showing how a schema can be created with a ```meta``` property. Probably the link from ```assignEntity``` to the tests should be extended to include the new example.

Feedback is welcome, thanks.

frankwallis:
If anyone has time to review this it would be appreciated, thanks

paularmstrong:
@frankwallis can you merge the latest master into your branch?
@gaearon will have final say on merging this.

frankwallis:
Thanks, I have merged the latest changes from master.

frankwallis:
Thank you!

paularmstrong:
This will be out in the next build. Likely 2.1.0 today or tomorrow.

bstro:
Man, this just made my life so much easier! A belated thank you to @paularmstrong and @frankwallis !

Issue #96, opened 2016-04-23
Different key between response and normalized response


exciton:
I'd like to be able to define a schema which uses a different key from what the server sends for the list of IDs. I'm fairly sure this is not currently possible.

E.g. server response like this:
```javascript
{
  id: 1,
  categories: [{ id: 10, name: 'Test'}]
}
```
into a normalized response like this
```javascript
{ result: 1,
  entities { 
    events {
      1: { category_ids: [10] }
    },
    categories {
      10: { name: 'Test' }
    }
  }
}
```
Note that the server returned a categories, but I want to represent it as category_ids in the normalized response. 

Currently I have a schema like this:
```javascript
const category = new Schema('categories')
const event = new Schema('events')
event.define({ categories: arrayOf(category) })
```

Is it possible to add an option to specify what key to use when assigning the normalized data? I don't believe the assignEntity function can do this (since it can't be set on a per-schema basis). schemaAttribute also isn't right. 
Perhaps the best way would be to allow assignEntity to be used in arrayOf etc.

e.g. 
```javascript
event.define({ 
  categories: arrayOf(category, { assignEntity: (obj, key, val) => obj['category_ids'] = val }) 
})
```

joewood:
How would this work for a many-to-one relationship? For example, articles to an auhor if you wanted the result (as above) to use `author_id` on `article`. Currently there's no way to pass an **options** argument:

```javascript
article.define({
  author: user,
});
```

Maybe if we had a `oneOf` function then we could pass an **options** object to it:
```javascript
article.define({
  author: oneOf(user,{ assignEntity: (obj,key,val)=> obj["author_id"]=val; })
});
```

frankwallis:
I think that https://github.com/paularmstrong/normalizr/pull/89 will solve this as you will be able to specify the new id on the schema and override assignEntity.

joewood:
@frankwallis how do you specify `assignEntity` as an option for a many-to-one relationship as above?

frankwallis:
```assignEntity``` is specified on the ```normalize``` call, and called for each entity which is normalized

exciton:
I think that #89 doesn't solve my case where I specifically want to specify assignEntities inside the schema definition. This is much more useful than specifying it when calling normalize since you don't need to know the details of the full schema and the code where normalize is called doesn't need to know anything about the specific schema or is normalizing. 

The assignEntities function is much more likely to be specific to a schema (or sub-schema) than it is to a normalize call.

[...]

PR #97, opened 2016-04-23
Add assignEntity option to IterableSchema

exciton:
Possibly this should also be added to other schemas, but this is a proof
of concept for now.

This PR refers to #96

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 .DS_Store
 node_modules
 dist
-lib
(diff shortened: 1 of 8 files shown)


paularmstrong:
I was recently thinking about doing something similar. 👍 for the proof of concept. I'd love to see this finalized!

Also, no need to commit the built files. Those will get created when we publish.

paularmstrong:
@exciton I put another implementation of this in #104. Also added tests that satisfy your use-case, I believe.

exciton:
@paularmstrong looks good!

paularmstrong:
Using #104

PR #104, opened 2016-05-11
Allow EntitySchema-specific assignEntity fns

paularmstrong:
This allows custom `assignEntity` functions on `EntitySchema`s.

**Problem:**

When dealing with many endpoints and normalizing across a lot of different schema, a single `assignEntity` function needs too many checks to determine whether it should have different behavior for particular keys. Often times, multiple schema will have the same keys but not need to be assigned the same way.

**Solution:**

Add an `assignEntity` option to `Schema`/`EntitySchema` that runs after the main `assignEntity` option passed to `normalize`.

Alternative to #97

### DIFF
--- a/src/EntitySchema.js
+++ b/src/EntitySchema.js
@@ -5,11 +5,16 @@ export default class EntitySchema {
     }
 
     this._key = key;
+    this._assignEntity = options.assignEntity;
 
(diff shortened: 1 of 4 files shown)


paularmstrong:
Merged manually.

PR #109, opened 2016-05-13
Quote all IDs, remove valuesOf, more thorough explanation for how merging works

wyuenho:
1. Updated README to reflect the changes with return string indices in normalized data.
2. Removed `valuesOf` and updated `arraysOf` to say they are essentially the same
2. More thorough explanation for how merging works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,22 +51,22 @@ can be normalized to
 
 ```javascript
 {
-  result: [1, 2],
+  result: ["1", "2"],
   entities: {
     articles: {
-      1: {
+      "1": {
         id: 1,
         title: 'Some Article',
         author: 1
       },
-      2: {
+      "2": {
         id: 2,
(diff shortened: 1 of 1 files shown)


paularmstrong:
Thanks for the contribution! At this time, I'm working on #105 to completely rewrite the documentation. It's a bit hard to follow at the moment, so I'd like to get it all taken care of in one big rewrite than try to piece things together. I hope my closing this in favor of some bigger work doesn't discourage you from further PRs on normalizr.

Also, a couple of nits/problems:

1. IDs are not converted to strings automatically in any way. If a number is provided, the number will be used. This may also need to be clarified.
2. `valuesOf` operates a bit differently than `arrayOf`. `valuesOf` uses the values of an object, whereas `arrayOf` expects an array to be given. I'll make sure this is a bit more explicit in #105

PR #117, opened 2016-05-26
Possibility to provide default values for the entity

### LINKED ISSUE #114: Possibility to provide default values to a model
# Problem

Sometimes it is useful to hydrate data with default values, when they are not provided in the server response. 

**Input**

It could look like this:

``` js
const item = new Schema('items', {defaults: {hello: 1}});
normalize([{id: 1, foo: 'bar'}, {id: 2, fuz: 'baz'}], arrayOf(item));
```

**Output**

``` js
{
  result: [1, 2],
  entities: {
    items: {
      1: {id: 1, foo: 'bar', hello: 1},
      1: {id: 2, fuz: 'baz', hello: 1}
    }
  }
}
```

smashercosmo:
This PR fixes #114.

- [x] Tests updated
- [x] Readme updated
- [x] Getting schemaAssignEntity is moved out of the loop for performance reasons

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -11,10 +11,11 @@ function defaultAssignEntity(normalized, key, entity) {
 function visitObject(obj, schema, bag, options) {
   const { assignEntity = defaultAssignEntity } = options;
 
-  let normalized = {};
+  const defaults = schema && schema.getDefaults && schema.getDefaults();
+  const schemaAssignEntity = schema && schema.getAssignEntity && schema.getAssignEntity();
+  let normalized = isObject(defaults) ? { ...defaults } : {};
   for (let key in obj) {
     if (obj.hasOwnProperty(key)) {
-      const schemaAssignEntity = schema && schema.getAssignEntity && schema.getAssignEntity();
       const entity = visit(obj[key], schema[key], bag, options);
       assignEntity.call(null, normalized, key, entity, obj, schema);
       if (schemaAssignEntity) {
--- a/README.md
+++ b/README.md
@@ -262,6 +262,9 @@ const article = new Schema('articles', { assignEntity: function (output, key, va
     obj[key] = value;
   }
 }})
+
+// You can specify default values for the entity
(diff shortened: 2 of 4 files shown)


paularmstrong:
Looks good! Just one small issue — the following test will fail:

```js
  it('does not overwrite the default', function () {
    var article = new Schema('articles', {defaults: {isFavorite: false}}),
      input;

    input = {
      id: 1
    };

    Object.freeze(input);

    normalize({ id: 2, title: 'foo' }, article);

    normalize(input, article).should.eql({
      result: 1,
      entities: {
        articles: {
          1: {
            id: 1,
            title: 'Some Article',
            isFavorite: false
          }
        }
      }
    });
  });
```

src/index.js
```diff
@@ -11,10 +11,11 @@ function defaultAssignEntity(normalized, key, entity) {
 function visitObject(obj, schema, bag, options) {
   const { assignEntity = defaultAssignEntity } = options;
 
-  let normalized = {};
+  const defaults = schema && schema.getDefaults && schema.getDefaults();
+  const schemaAssignEntity = schema && schema.getAssignEntity && schema.getAssignEntity();
+  let normalized = isObject(defaults) ? defaults : {};
```
paularmstrong: To fix the test I noted, I believe you just need to make a new object that includes the default values:

```js
let normalized = isObject(defaults) ? { ...defaults } : {};
// or
let normalized = isObject(defaults) ? Object.assign({}, defaults) : {};
```

smashercosmo: Good catch. My bad (( didn't think about object reference. I'll go with spread, because as I understand Object.assign should be polyfilled.

smashercosmo:
PR updated

paularmstrong:
@smashercosmo (sorry, been on holiday) Looks good. I'll mark it to be merged and in the next release. I'd like to hold off on merging for now so that the documentation doesn't add any confusion as to why it's not available yet.

smashercosmo:
Any news on this?

Issue #116, opened 2016-05-26
Different attribute values for same entity


lmuntaner:
# Problem

We have a situation where the same entity has different values of a specific attribute depending on where that entity comes from.

**Input**

Following your example with articles, this would be the input:

```js
var article = new Schema('articles'),
        tutorial = new Schema('tutorials'),
        magazine = new Schema('magazine'),
        articleOrTutorial = { articles: article, tutorials: tutorial },
        input;

magazine.define({
  contents: arrayOf(articleOrTutorial, { schemaAttribute: 'type' } )
});

input = [
  {
    id: 1,
    name: 'my magazine',
    contents: [{
        id: 1,
        type: 'articles',
        title: 'Some Article',
        releaseDate: '01/01/2016'
      }, {
        id: 2,
        type: 'articles',
        title: 'Another Article',
        releaseDate: '10/01/2016'
      }, {
        id: 1,
        type: 'tutorials',
        title: 'Some tutorial',
        releaseDate: '10/01/2016'
      }]
    },
    {
      id: 2,
      name: 'your magazine',
      contents: [{
        id: 1,
        type: 'articles',
        title: 'Some Article',
        releaseDate: '15/01/2016'
      }, {
        id: 2,
        type: 'articles',
        title: 'Another Article',
        releaseDate: '25/01/2016'
      }, {
        id: 1,
        type: 'tutorials',
        title: 'Some tutorial',
        releaseDate: '25/01/2016'
      }]
    }
  };

output = normalize(input, arrayOf(magazine);
```

**Output**

When we normalize the input, we get merge conflict with the `releaseDate`. We lose the information regarding the specifics of that article or tutorial related to its magazine. The articles are the same and I want to keep them as the same entity. But I'd like to also keep the value that is specific of that article related with the magazine.

I'd like to get maybe something like this:
```js
output = {
  magazines: {
    1:  {
      id: 1,
      name: 'my magazine',
      contents: [{
          id: 1,
          schema: 'articles',
          releaseDate: '01/01/2016'
        }, {
          id: 2,
          schema: 'articles',
          releaseDate: '10/01/2016'
        }, {
          id: 1,
          schema: 'tutorials',
          releaseDate: '10/01/2016'
        }
      ]
    },
    2: {
      id: 2,
      name: 'your magazine',
      contents: [{
          id: 1,
          schema: 'articles',
          releaseDate: '15/01/2016'
        }, {
          id: 2,
          schema: 'articles',
[...]

PR #118, opened 2016-05-26
Polymorphic item creator

lmuntaner:
# Problem

Related to [issue#116](https://github.com/paularmstrong/normalizr/issues/116).

The same entity has different values of a specific attribute depending on where that entity comes from.

When we normalize the input, we get merge conflict with the releaseDate. We lose the information regarding the specifics of that article or tutorial related to its magazine. The articles are the same and I want to keep them as the same entity. But I'd like to also keep the value that is specific of that article related with the magazine.

# Solution

Provide a `polymorphicItem` key with a function on polymorphicMaps. That function will add more keys to the reference object.
```js
  function polymorphicItem(item) {
    return { releaseDate: item.releaseDate };
  }

  magazine.define({
    contents: arrayOf(articleOrTutorial, { schemaAttribute: 'type', polymorphicItem: polymorphicItem } )
  });
```

I hope this is clear enough 😺

### DIFF
--- a/src/UnionSchema.js
+++ b/src/UnionSchema.js
@@ -15,11 +15,7 @@ export default class UnionSchema {
     const schemaAttribute = options.schemaAttribute;
     this._getSchema = typeof schemaAttribute === 'function' ? schemaAttribute : x => x[schemaAttribute];
 
-    function defaultPolymorphicItem(item) {
-      return { schema: this._getSchema(item) };
-    }
-
-    this._getSchemaKeys = options.polymorphicItem || defaultPolymorphicItem;
+    this._getPolimorphicItem = options.polymorphicItem || ( () => ({}) );
   }
 
   getItemSchema() {
@@ -30,7 +26,7 @@ export default class UnionSchema {
     return this._getSchema(item);
   }
 
-  getSchemaKeys(item) {
-    return this._getSchemaKeys(item);
+  getPolimorphicItem(item) {
+    return this._getPolimorphicItem(item);
   }
 }
--- a/src/UnionSchema.js
+++ b/src/UnionSchema.js
@@ -15,7 +15,7 @@ export default class UnionSchema {
     const schemaAttribute = options.schemaAttribute;
     this._getSchema = typeof schemaAttribute === 'function' ? schemaAttribute : x => x[schemaAttribute];
 
-    this._getPolimorphicItem = options.polymorphicItem || ( () => ({}) );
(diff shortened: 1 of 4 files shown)


src/UnionSchema.js
```diff
@@ -14,14 +14,19 @@ export default class UnionSchema {
 
     const schemaAttribute = options.schemaAttribute;
     this._getSchema = typeof schemaAttribute === 'function' ? schemaAttribute : x => x[schemaAttribute];
+
+    this._getPolimorphicItem = options.polymorphicItem || ( () => ({}) );
```
paularmstrong: nit: `polymorphic` is the more common spelling.

src/index.js
```diff
@@ -31,9 +32,10 @@ function defaultMapper(iterableSchema, itemSchema, bag, options) {
 
 function polymorphicMapper(iterableSchema, itemSchema, bag, options) {
   return (obj) => {
+    const polimorphicItemKeys = iterableSchema.getPolimorphicItem(obj);
     const schemaKey = iterableSchema.getSchemaKey(obj);
     const result = visit(obj, itemSchema[schemaKey], bag, options);
-    return { id: result, schema: schemaKey };
+    return merge({ id: result, schema: schemaKey }, polimorphicItemKeys);
```
paularmstrong: Can do this without `lodash/merge`:

```js
return { ...polymorphicItemKeys, id: result, schema: schemaKey };
```

src/UnionSchema.js
lmuntaner: sure! Like `options.polimorphic` and `this._getPolimorphic`??

src/index.js
lmuntaner: I will

src/UnionSchema.js
lmuntaner: haha, sorry, I just saw the typo!

lmuntaner:
Good morning @paularmstrong ,

Do you need anything else? Or do you suggest any other change? I can update the Readme as well if you want to. We would really appreciate this feature on our project.

Thanks! 😸

paularmstrong:
I'm still not convinced that this is the right thing to do. Can we continue the discussion on #116?

lmuntaner:
sure

paularmstrong:
Closing in favor of solution here: [#116](https://github.com/paularmstrong/normalizr/issues/125 116#issuecomment-228738326)

Issue #130, opened 2016-07-01
Allow dynamic schemaAttribute for children by checking a key on the parent-level


jgutta:
In this example from your read me:

``` js
const group = new Schema('groups');
const user = new Schema('users');

// a member can be either a user or a group
const member = {
  users: user,
  groups: group
};

// You can specify the name of the attribute that determines the schema
group.define({
  owner: unionOf(member, { schemaAttribute: 'type' })
});
```

is normalizr looking for the schemaAttribute in the "member" object, or the "group" object? If it is looking at the "member"  object, is there a way to make it check "group"  instead?

EDIT: To answer half of my own question, Normalizr looks for the schemaAttribute in the "member" object. I am still working on a way to check the parent, any advice in that direction would be appreciated!

fxck:
I've got the exact same problem.

fxck:
@paularmstrong @gaearon any advice? basically my by data structure looks like this 

```
{
  // ...
  schema: 'foo',
  data: { /* ... */ }
}
```

and I want `data` to be a schema of something depending on what's inside the `schema`, the closest I got was by using `unionOf`(which doesn't even make that much sense here I think), but what get's passed to `schemaAttribute` is what's inside of `data`, not its parent..

ideal would be if this worked

```typescript

const somethingSchema = new Schema('something');
const fooSchema = new Schema('foo');

const schemaMap = {
  foo: fooSchema
};
something.define({
  data: (entity) => schemaMap[entity.schema]
});
```

paularmstrong:
This is not available yet. I'm not sure how easy it'd be, either... at a cursory glance, that looks like a pretty big overhaul. If you'd like to try it out and open a PR, I'd be happy to look.

fxck:
Damn that's too bad @paularmstrong I tried looking into the source code to see if there was a simple way to implement it prior to writing down the comment, and I couldn't quite figure out where to even begin..

fxck:
So I gave it a shot, I think the only change in the end was at this line https://github.com/paularmstrong/normalizr/blob/master/src/index.js#L18

where instead of just passing it `schema[key]`, I do `const resolvedSchema = typeof schema[key] === 'function' ? schema[key].call(null, obj) : schema[key];` instead.. it seems to work, what do you think? 

I'll try to write some tests

fxck:
https://github.com/paularmstrong/normalizr/pull/132

jgutta:
[...]

PR #132, opened 2016-07-06
(feat) add ability to dynamically set nested schema type

fxck:
# Problem

There is currently no way to set nested schema type depending on some value from the parent schema. https://github.com/paularmstrong/normalizr/issues/130#issuecomment-230622131

# Solution

This PR adds ability to pass a function to `define`s object values, for example 

```javascript
var articleSchema = new Schema('article');
var mediaSchema = new Schema('media');
var listsSchema = new Schema('lists');

var schemaMap = {
  media: mediaSchema,
  lists: listsSchema
};

articleSchema.define({
  foo: (article) => schemaMap[article.type]
});
```

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -877,40 +877,48 @@ describe('normalizr', function () {
     });
   });
 
-  it('can normalize nested entities', function () {
-    var article = new Schema('articles'),
-        user = new Schema('users'),
+  it('can normalize nested entity using property from parent', function () {
+    var linkablesSchema = new Schema('linkables'),
+        mediaSchema = new Schema('media'),
+        listsSchema = new Schema('lists'),
         input;
 
-    article.define({
-      author: user
+    var schemaMap = {
+      media: mediaSchema,
+      lists: listsSchema
+    };
+
+    linkablesSchema.define({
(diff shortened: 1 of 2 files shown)


fxck:
meh.. travis is failing because

```
npm ERR! fetch failed https://registry.npmjs.org/babel-plugin-transform-object-rest-spread/-/babel-plugin-transform-object-rest-spread-6.8.0.tgz
```

test/index.js
```diff
@@ -877,40 +877,48 @@ describe('normalizr', function () {
     });
   });
 
-  it('can normalize nested entities', function () {
```
paularmstrong: Please don't change this test, just add another. Right now, the tests also act as documentation for what can be done.

fxck: Oups, my mistake, thought I was copy pasting and apparently rewrote the test instead, let me fix that!

fxck:
@paularmstrong is there anything else I need to do?

paularmstrong:
@fxck Can you update so this will merge cleanly?

fxck:
closing in favour of https://github.com/paularmstrong/normalizr/pull/159

PR #139, opened 2016-07-26
Feature - Opportunity to add properties after assignEntity

ajwhite:
# Problem

`assignEntity` is great for transforming individual values for particular keys, or using an individual key and value to derive a new one. But what if there's a value that depends on multiple keys and values, or after the model has been fully processed by `assignEntity`? 

I have a situation where there's a handful of fields that are required to derive a value for a particular key on the entity. The best place for this to happen would be _after_ the individual properties have been processed by `assignEntity`.

A simple example would be to compute a `hash` based on a set of fields for efficient comparisons downstream. 

Currently the only way to do this would be to check if the object contains the necessary keys in `assignEntity`:

```js
assignEntity (obj, key, value) {
  if (obj.a && obj.b && obj.c && obj.d) {
    // if these fields are needed to perform a calculation, the only way to know
    // that they have been processed is by checking their truthiness together.
    
    obj.hash = hashingFn(obj.a, obj.b, obj.c, obj.d);
  }
}
```

# Solution

A method, similar to `assignEntity`, is added to the `Schema` `options`, or to the `options` passed to `normalize`. When an entity is visited by `visitObject`, after it has completed the `assignEntity` procedure it will call `assignEntityCompleted` with the `normalized` object reference. This opens up the opportunity to apply any finishing touches to the entity.

```js
assignEntityCompleted: function (obj) {
  obj.hash = hashingFn(obj.a, obj.b, obj.c, obj.d);
}
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,6 +4,9 @@ export type ExtractAttribute =
 export type AssignEntity =
   (output: any, key: string, value: any, input: any, schema: SchemaValue) => void;
 
+export type AssignEntityCompleted =
+  (output: any) => void;
+
 export type MergeIntoEntity =
   (entityA: any, entityB: any, entityKey: string) => void;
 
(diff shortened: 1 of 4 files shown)


paularmstrong:
I'm going to pass on this. It's easy enough to do on your own after the entire data has been normalized.

PR #143, opened 2016-08-08
return class instance from Schema.define

just-boris:
After this change it will be possible to write the following:

```js
const post = new Schema('posts').define({
    comments: arrayOf(comment)
});
```

It is better to read when you have a lot of shemas definition and don't have so much overhead that was in #46.

This is not intended to merge as is, this more a question why it wasn't done in first place.

### DIFF
--- a/src/EntitySchema.js
+++ b/src/EntitySchema.js
@@ -47,5 +47,6 @@ export default class EntitySchema {
         this[key] = nestedSchema[key];
       }
     }
+    return this;
   }
 }


paularmstrong:
I understand why you think this makes things simpler, but it changes the expectations. Once you see chaining happen once, you'd expect it to work everywhere (thanks, jQuery). So with this, I'd assume I could chain multiple times... `.define(...).define(...)`, etc.

Thanks for opening a pull request, but I'm going to pass on merging this. You could always create a simple wrapper in your own project that would handle this the same way!

PR #145, opened 2016-08-12
add defaults typings

Kestutis:
# Problem

defaults attribute is not preset in SchemaOptions type

**Input**

```js
new Schema('data', {
    idAttribute: 'ID',
    defaults : { defaultValue: 1 }
});
```
**Output**

I'm getting following error from TypeScript
`Object literal may only specify known properties, and 'defaults' does not exist in type ...`

**Solution** 

Add `defaults?:any` to type script definition file on SchemaOptions type:
```js
export type SchemaOptions = {
  idAttribute?: string | ExtractAttribute;
  meta?: any;
  assignEntity?: AssignEntity;
  defaults?: any;
}
```

also add `getDefaults(): any;` to Schema class.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -11,6 +11,7 @@ export type SchemaOptions = {
   idAttribute?: string | ExtractAttribute;
   meta?: any;
   assignEntity?: AssignEntity;
+  defaults?: any;
 }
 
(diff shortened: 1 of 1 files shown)


metanav:
+1
ping @paularmstrong

PR #73, opened 2016-02-26
can now normalize iterables keyed with their id

adjohu:

PR #155, opened 2016-09-23
Can normalize entities keyed with their ID

timhwang21:
# Problem

It is hard to deal with collections keyed by ID without an ID in the entity itself
# Solution

Refactored #73 by to pass test. Allows `idAttribute` to be passed a `key` with which to normalize IDs on. Also fixed some instances where `collectionKey` wasn't being passed to `visit()`, and added documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,6 +248,12 @@ const article = new Schema('articles', { idAttribute: 'slug' });
 function generateSlug(entity) { /* ... */ }
 const article = new Schema('articles', { idAttribute: generateSlug });
 
+// The function is passed both the current value and its key
+// This is helpful in occasions where your data is keyed by id, but doesn't contain id inside the value
+const keyedByIdJson = { 1: { ... }, 2: { ... }, 3: { ... } };
+function generateSlug(entity, id) { return id; }
+const article = new Schema('articles', { idAttribute: generateSlug });
+
 // You can also specify meta properties to be used for customizing the output in assignEntity (see below)
 const article = new Schema('articles', { idAttribute: 'slug', meta: { removeProps: ['publisher'] }});
 
--- a/src/index.js
+++ b/src/index.js
@@ -8,7 +8,7 @@ function defaultAssignEntity(normalized, key, entity) {
   normalized[key] = entity;
 }
 
-function visitObject(obj, schema, bag, options) {
+function visitObject(obj, schema, bag, options, collectionKey) {
   const { assignEntity = defaultAssignEntity } = options;
 
   const defaults = schema && schema.getDefaults && schema.getDefaults();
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -269,7 +275,7 @@ const article = new Schema('articles', { defaults: { likes: 0 } });
 
 ### `Schema.prototype.define(nestedSchema)`
 
-Lets you specify relationships between different entities.
```
paularmstrong: Please preserve trailing whitespace in markdown. Two spaces at the end of a line indicates a line break.

src/index.js
```diff
@@ -90,28 +90,35 @@ function visitEntity(entity, entitySchema, bag, options) {
...
     return visitIterable(obj, schema, bag, options);
   } else if (schema instanceof UnionSchema) {
     return visitUnion(obj, schema, bag, options);
   } else {
-    return visitObject(obj, schema, bag, options);
+    return visitObject(obj, schema, bag, options, collectionKey);
   }
 }
 
+function normalizeResult(result) {
```
paularmstrong: Why do we need this function? When is the result coming back as an object?

test/index.js
```diff
@@ -1902,4 +1902,37 @@ describe('normalizr', function () {
...
+        name: 'Adam'
+      },
+      4: {
+        name: 'Jeremy'
+      }
+    };
+
+    Object.freeze(input);
+
+    normalize(input, arrayOf(user)).should.eql({
```
paularmstrong: `arrayOf` seems like a poor use for an object. Would it simplify things to add a `keyedObject` function that assumes your `ID` is the `key` for each entry?

timhwang21: Well, right now the functionality that the previous author wrote is baked into `visitIterable()` (line 48). So like `valuesOf()`, this function would create a `new IterableSchema` anyways. I do agree that `valuesOf()` makes more sense, and perhaps `keyedValuesOf()` more. What do you think about a `keyedValuesOf()` that's yet another alias of `arrayOf()`? Not sure if that's unnecessary API complication, though.

src/index.js
timhwang21: Result comes back as an object because the new functionality is added to an iterable schema (`arrayOf(foo)` or `keyedObjectOf(foo)`, as mentioned above). Thus, `visit()` returns the result of `visitIterable()`, which is an object mapping keys to the result of `curriedItemMapper(obj[key], key)`. 

When we use the function `(obj, key) => key`, this ultimately returns an object something like `{ 1: '1', 2: '2', 3: '3' }` (whereas something like `obj => obj` would return `{ 1: foo, 2: bar, 3: baz }` or something). Here, the values of the object don't represent objects, so we convert into an array of keys after double checking that all the keys do equal all their values.

specialunderwear: It does happen that the result can not be normalized, if you specify idAttribute for example, your key will not be your id, and the result can not be further normalized. So every time you don;t want the key to be the ID, but you still need it to computed the id, you will not get and array, but the object stays.

paularmstrong:
@timhwang21 Can you fix the merge conflicts when you have a chance? I'd like to get this merged and deployed this week.

timhwang21:
@paularmstrong done!

PR #158, opened 2016-10-15
Update EntitySchema's define method

canvaskisa:
# Problem

The problem is that we cannot easily create schemas by one-liner:
```js
const postSchema = new Schema('posts').define({user: userSchema});
```
It is not a big deal, really, but a tasty feature :icecream:.

# Solution
Simply return `this` from `EntitySchema` `define` method.


canvaskisa:
Also removes weird trailing whitespace on 39 line.

paularmstrong:
Please refer to #143

PR #161, opened 2016-10-19
Pass parent object to idAttribute.

### LINKED ISSUE #160: Deeply Nested Schema Question
# Problem

I am trying to normalize a deep-ish array of objects. The issue is that the "key" I want to use is at one level, and the data I am able to normalize is at a deeper nesting level. I want to key off of "sku" but only put "basic" into the entity cache.

**Input**

``` js
/* The approximate shape of the JSON data is. I have pruned it down for clarity
{
  "totalCount":25,
  "orders":[
    {
      "id":"2031948139",
      "version":10,
      "basic":{...},
      "details":{...},
      "items":[
        {
          "orderData":{...},
          "product":{
            "sku":"1025879",
            "upcs":[...],
            "basic":{
              "name":"Product 1",
              ...
              "image":{...},
            },
            "store":{
              "price":{
                "list":5.98,
                "salesUnitOfMeasure":"OZ"
              }
            }
          }
        },
        {
          "orderData":{...},
          "product":{
            "sku":"3000691800",
            "upcs":[...],
            "basic":{
              "name":"Product 2",
              ...
              "image":{...},
            },
            "store":{
              "price":{
                "list":2.98,
                "salesUnitOfMeasure":"OZ"
              }
            }
          }
        },
        ...
      ]
    },
    ...
  ]
}
*/

const order = new Schema('Order');
const product = new Schema('Product', { assignEntity: function (output, key, value, input) {
  if (key === 'sku') {
    output[value] = input.basic
  }
}});

order.define({
  items: arrayOf({product})
});

const result = normalize(res.value, {
  orders: arrayOf(order)
});

```

**Output**

``` js
{
  entities: {
    Order: {
      2031948139: {
        items: [{ 
          orderData: {...},
          product: {
            basic: '1025879',
            sku: '1025879',
            "store":{
              "price":{
                "list":5.98,
                "salesUnitOfMeasure":"OZ"
              }
            }
          }
        }, {
          orderData: {...},
          product: {
            basic: '3000691800'
            sku: '3000691800'
            "store":{
              "price":{
                "list":2.98,
                "salesUnitOfMeasure":"OZ"
              }
            }
          }
        }]
      }
    }
    Product: {
      '1025879': {}
      '3000691800': {}
    }
  }
}
```

Basically I want to only normalize the `basic` product information but key it off the sku value. My attempt at using `assignEntity` doesn't seem to work like I think it does. It appears that `assignEntity` mutates `output` regardless of what happens inside the internals. Any suggestions would be appreciated.

**Update 10/18/2016** - Think what would be ideal is to have access to the parent entity in the generateId function. That would easily solve this issue.

arb:
# Problem

Provide a mechanism to use a parent object data in generating entity id.
# Solution

Pass the `last` object through all of the traversal methods and make sure it gets passed specifically to `visitEntity` as that is where `getId` is called and pass the parent object into that function. 

Closes #160

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -8,7 +8,7 @@ function defaultAssignEntity(normalized, key, entity) {
   normalized[key] = entity;
 }
 
-function visitObject(obj, schema, bag, options) {
+function visitObject(obj, schema, parentObj, bag, options) {
   const { assignEntity = defaultAssignEntity } = options;
 
   const defaults = schema && schema.getDefaults && schema.getDefaults();
@@ -16,7 +16,7 @@ function visitObject(obj, schema, bag, options) {
   let normalized = isObject(defaults) ? { ...defaults } : {};
   for (let key in obj) {
     if (obj.hasOwnProperty(key)) {
-      const entity = visit(obj[key], schema[key], bag, options);
+      const entity = visit(obj[key], schema[key], obj, bag, options);
       assignEntity.call(null, normalized, key, entity, obj, schema);
       if (schemaAssignEntity) {
         schemaAssignEntity.call(null, normalized, key, entity, obj, schema);
@@ -26,21 +26,21 @@ function visitObject(obj, schema, bag, options) {
   return normalized;
 }
 
-function defaultMapper(iterableSchema, itemSchema, bag, options) {
-  return (obj) => visit(obj, itemSchema, bag, options);
+function defaultMapper(iterableSchema, itemSchema, parentObj, bag, options) {
+  return (obj) => visit(obj, itemSchema, parentObj, bag, options);
 }
 
-function polymorphicMapper(iterableSchema, itemSchema, bag, options) {
+function polymorphicMapper(iterableSchema, itemSchema, parentObj, bag, options) {
   return (obj) => {
     const schemaKey = iterableSchema.getSchemaKey(obj);
-    const result = visit(obj, itemSchema[schemaKey], bag, options);
+    const result = visit(obj, itemSchema[schemaKey], parentObj, bag, options);
     return { id: result, schema: schemaKey };
   };
 }
 
-function visitIterable(obj, iterableSchema, bag, options) {
+function visitIterable(obj, iterableSchema, parentObj, bag, options) {
   const itemSchema = iterableSchema.getItemSchema();
-  const curriedItemMapper = defaultMapper(iterableSchema, itemSchema, bag, options);
+  const curriedItemMapper = defaultMapper(iterableSchema, itemSchema, parentObj, bag, options);
 
   if (Array.isArray(obj)) {
     return obj.map(curriedItemMapper);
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -8,15 +8,15 @@ function defaultAssignEntity(normalized, key, entity) {
...
+function visitObject(obj, schema, bag, options, last) {
   const { assignEntity = defaultAssignEntity } = options;
 
   const defaults = schema && schema.getDefaults && schema.getDefaults();
   const schemaAssignEntity = schema && schema.getAssignEntity && schema.getAssignEntity();
   let normalized = isObject(defaults) ? { ...defaults } : {};
   for (let key in obj) {
     if (obj.hasOwnProperty(key)) {
-      const entity = visit(obj[key], schema[key], bag, options);
+      const entity = visit(obj[key], schema[key], bag, options, obj);
```
paularmstrong: `parentObject` instead of `last` would be a lot more clear.

src/index.js
```diff
@@ -8,15 +8,15 @@ function defaultAssignEntity(normalized, key, entity) {
   normalized[key] = entity;
 }
 
-function visitObject(obj, schema, bag, options) {
+function visitObject(obj, schema, bag, options, last) {
```
paularmstrong: All of these function signatures are starting to get a little long. I'd like to at least see the order make more sense (optional objects last)

```
function visitObject(obj, schema, parentObj, bag, options) { ... }
```

arb:
@paularmstrong thanks for the feedback. I have addressed all of your issues I believe.

paularmstrong:
@arb I'm getting test failures after merging with master. Can you update when you have a minute so I can get a new version deployed this week?

arb:
@paularmstrong are you sure that's the way to go? I feel like 7d527b64c6d5f19486ce9d4c2225b56061e36744 is a subset of the feature I'm trying to add here. Additionally, if I merge in the changes from 7d527b64c6d5f19486ce9d4c2225b56061e36744 now those internal functions with 5 arguments will now have 6 arguments.

arb:
@paularmstrong any update on my question above? I'd like to get this merged in so we can use it in our project.

paularmstrong:
If I can find some time over the next week, I'll re-evaluate what we're doing. Might end up making some significant changes to the internal API, since things have gotten more complicated than I realized.

paularmstrong:
The more I look at this, the less it makes sense. Maybe you can help me understand why this would be necessary.

From the test example, it looks like you have a strict 1-to-1 from your parent object to the `meta` object. Why would you need to normalize the `meta` out as its own entity? From my own experience with redux applications, it's better to just leave this information in place. That way, you always have this reference you're looking for.

arb:
It is possible my use case is slightly specialized...

But the situation I'm in is that I'm getting data back that looks like this

```json
product:{
 id: 123,
 basic: {},
 store:{}
}
```

The stuff in `product.basic` is pretty static so it's safe to store in an entity cache. The issue is, the stuff inside `basic` does not have a unique identifying value. The `id` is up a level. What I *want* to do, is put the data under `product.basic` into an entity cache somewhere, and leave `store` alone as it changes too much to get any value of it. 

Currently, there is no way with my existing data model to do this.

paularmstrong:
Are you using redux? Your reducer should be handling what is and is not stored.

arb:
Yes to redux. This is before the fetch data even hits any reducers. I want to do the normalization right after the data has been fetched, not at the reducer level.

paularmstrong:
This is now possible in v3.0.0-alpha.5! When passing a function to the [idAttribute (docs)](https://github.com/paularmstrong/normalizr/blob/c85dfa0d7924fe2fbcbf526ffd5cbd0a3819f672/docs/api.md#entitykey-definition---options--), you're given the `value`, `parent`, and `key` arguments. I hope this new version satisfies what you need!

I'm going to close this PR in favor of Normalizr v3. Thank you so much for taking the time to contribute to open source and your interest in normalizr!

arb:
Yeah that looks like it should do it. Thanks.

PR #173, opened 2016-11-08
Added schema option for assigning the parent id on the child entity

### LINKED ISSUE #172: Set field on child entity based on parent entity on normalization
# Problem

Is there anyway to set a field on the child entity based on the parent entity? I.e, if we have

```
[{
  id: 1,
  title: 'Some Article',
  comment: {
    id: 109,
    content: 'Hai'
  }
}, {
  id: 2,
  title: 'Other Article',
  comment: {
    id: 110,
    content: 'Gee'
  }
}]
```

Can we somehow get 

```
{
  result: [1, 2],
  entities: {
    articles: {
      1: {
        id: 1,
        title: 'Some Article',
        comments: [109]
      },
      2: {
        id: 2,
        title: 'Other Article',
        comments: [110]
      }
    },
    comments: {
      109: {
        id: 109,
        content: 'Hai'
        article: 1,
      },
      110: {
        id: 110,
        content: 'Gee'
        article: 2,
      }
    }
  }
}
```
?

Would this be a desirable feature otherwise?

poyannabati:
# Problem

Closes #172. Partially inspired by #135. 

Basically, we sometimes when normalizing, we would like to set the parent id on the child as well as setting the child ids on the parent. 

Example;

```
// INPUT 

[{
  id: 1,
  title: 'Some Article',
  comment: {
    id: 109,
    content: 'Hai'
  }
}, {
  id: 2,
  title: 'Other Article',
  comment: {
    id: 110,
    content: 'Gee'
  }
}]

// OUTPUT

{
  result: [1, 2],
  entities: {
    articles: {
      1: {
        id: 1,
        title: 'Some Article',
        comments: [109]
      },
      2: {
        id: 2,
        title: 'Other Article',
        comments: [110]
      }
    },
    comments: {
      109: {
        id: 109,
        content: 'Hai'
        article: 1,
      },
      110: {
        id: 110,
        content: 'Gee'
        article: 2,
      }
    }
  }
}
```

# Solution

When defining the schema, you can set assignParentId to true and this will result in the child entity getting the parent object's id set. 

# TODO
- [ ] Didn't cover the UnionSchema case. Let me know if that is a requirement to proceed. 
- [x] Add & Update Tests

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -1950,4 +1950,88 @@ describe('normalizr', function () {
     }).should.throw();
   });
 
+  it('can assign parent id on child entity', function () {
+    var article = new Schema('articles'),
+      comment = new Schema('comments', { assignParentId: true }),
+      input;
+
+    article.define({
+      comments: arrayOf(comment),
+    });
+
+    input = [{
+      id: 1,
+      title: 'Some Article',
+      comments: [{
+        id: 109,
+        content: 'Hai'
+      }]
+    }, {
+      id: 2,
+      title: 'Other Article',
+      comments: [{
+        id: 110,
+        content: 'Gee'
+      }]
+    }];
+
+    normalize(input, arrayOf(article)).should.eql({
+      result: [1, 2],
+      entities: {
+        articles: {
+          1: {
+            id: 1,
+            title: 'Some Article',
+            comments: [109]
+          },
+          2: {
+            id: 2,
+            title: 'Other Article',
+            comments: [110]
+          }
+        },
+        comments: {
+          109: {
+            id: 109,
+            content: 'Hai',
+            articles: 1,
+          },
+          110: {
+            id: 110,
+            content: 'Gee',
+            articles: 2,
+          }
+        }
+      }
+    });
+  });
+
+  it('can normalize parentless entities even when assignParentId is truthy', function () {
+    var comment = new Schema('comments', { assignParentId: true }),
+      input;
+
+    input = [{
+        id: 109,
+        content: 'Hai',
+      }
+    , { id: 110,
+        content: 'Gee',
+    }];
+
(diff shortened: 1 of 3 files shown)


test/index.js
```diff
@@ -1950,4 +1950,88 @@ describe('normalizr', function () {
     }).should.throw();
   });
 
+  it('can assign parent id on child entity', function () {
+    var article = new Schema('articles'),
+      comment = new Schema('comments', { assignParentId: true }),
```
paularmstrong: What happens if you try to do this and there are multiple parents?

poyannabati:
Filtering/finding is the backup option I've considered. However, I don't think it doesn't (double negative) make sense to include an option to allow for traversing the object graph in the opposite direction without requiring a search.

test/index.js
poyannabati: Didn't cover this case, but a suggestion is that when there are multiple parents the child is assigned an array of parent ids.

paularmstrong:
Does [this](https://github.com/paularmstrong/normalizr/blob/34e101d3433da4ce06346133b718439f34a7a12e/test/schemas/Entity.test.js#L31-L38) solve your problem?

paularmstrong:
I've added [an example](https://github.com/paularmstrong/normalizr/tree/c85dfa0d7924fe2fbcbf526ffd5cbd0a3819f672/examples/relationships) to the v3.0.0 branch to show how you might handle this manually. I think this may have to be more of a manual setup process for everyone similar to the example, as you'll always need to handle the `mergeStrategy` properly, since many-to-many relationships need attention when merging duplicates.

I'm going to close this PR in favor of that example and because I would like to freeze v2 in favor of future work going into v3. Thank you so much for taking the time to contribute to open source and your interest in normalizr!

PR #178, opened 2016-11-23
index.d.ts should be located in the same directory as index.js (build)

### LINKED ISSUE #146: index.d.ts should be in the same directory as index.js
Currently the `index.d.ts` is located in the package root. This is a problem, because it does not define an interface which is attached to a global object, but an interface which is exported. This cause for example problems in situation where imports for `Schema` & co. are auto-generated and the index.js file with the actual source is located in a different location.

akomm:
Fixes #146


package.json
```diff
@@ -44,6 +44,7 @@
     "babel-preset-es2015": "^6.18.0",
     "babel-preset-stage-1": "^6.16.0",
     "babel-register": "^6.18.0",
+    "copy-webpack-plugin": "^4.0.1",
```
paularmstrong: Please run  `yarn` to update the `yarn.lock` file.

akomm:
Note, that webpack2 now handles correctly misplaced `.d.ts` importing the `main` file from `package.json` out of the box. In case of this package is pointing to the correct source (`lib/index.js`) matching the `index.d.ts`.

zaycker:
Guys, we are using webpack@1 and TypeScript@1.8.x and build is failed. Typescript compiler can't find module.
Changing path of `.d.ts` file to `"typings": "lib/index.d.ts"` in `package.json` fixes problem.
Help :)

paularmstrong:
@zaycker I don't really have any experience with typescript and rely on the community for that. If you have a fix, please create a PR.

PR #183, opened 2016-12-08
#178 index.d.ts is in the same directory as index.js

zaycker:
# Problem

TypeScript cannot resolve module path while compiling. A recent change moved the location of index.js and index.d.ts and the entry point for index.js inside package.json. But updating the "typings" key in package.json was missed.

# Solution

Fill in right path to typings

# TODO

- [ ] Add & Update Tests


paularmstrong:
Merged and published v2.3.1

PR #191, opened 2016-12-21
Point typings to correct path

unindented:
# Problem

After installing `normalizr`, the `index.d.ts` seems to end up in the `lib` folder. The `typings` attribute in `package.json` points to `index.d.ts` instead of `lib/index.d.ts`, so the TypeScript compiler is not able to find it.

# Solution

Point to the correct path.

# TODO

*Fill in the following:*

- [ ] Add & Update Tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "2.3.0",
   "description": "Normalizes JSON according to schema for Redux and Flux applications",
   "main": "lib/index.js",
-  "typings": "index.d.ts",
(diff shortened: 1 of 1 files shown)


unindented:
Ah, this is a dupe of https://github.com/paularmstrong/normalizr/pull/183.

@paularmstrong could you merge one of the two and publish a minor version bump, just to fix the typings?

paularmstrong:
Merged and published #183

PR #192, opened 2016-12-27
Typings for normalizr v3

unindented:
WIP

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,10 +19,12 @@
   ],
   "files": [
     "dist/",
+    "index.d.ts",
     "LICENSE",
     "README.md"
   ],
   "main": "dist/index.js",
+  "typings": "index.d.ts",
   "scripts": {
     "build": "npm run clean && mkdirp dist && npm-run-all --parallel build:development build:production",
     "build:development": "NODE_ENV=development rollup -c",
@@ -56,6 +58,8 @@
     "rollup": "^0.37.0",
     "rollup-plugin-babel": "^2.7.1",
     "rollup-plugin-filesize": "^1.0.1",
-    "rollup-plugin-uglify": "^1.0.1"
+    "rollup-plugin-uglify": "^1.0.1",
+    "typescript": "^2.1.4",
+    "typescript-definition-tester": "0.0.5"
   }
 }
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -19,10 +19,12 @@
   ],
   "files": [
     "dist/",
+    "index.d.ts",
```
paularmstrong: I've gotten issues reported with problems of the typings file being in a different directory than the module itself. Can we put it in `src` and make sure it's copied to `dist/` and referenced there?

... Unless those issues are totally unfounded and there's something I'm missing?

unindented: I think if the `typings` field points to the right place it shouldn't matter, but I'll test everything out.

I wanted to avoid copying, as `cp` is *nix-specific, and I wanted to avoid adding a dependency like `ncp` just for that task.

paularmstrong:
Weird... I didn't mean to close this (not sure how I did)... and it won't let me reopen O_o

paularmstrong:
Ah... because I deleted the base branch... Changed to master. All good.

paularmstrong:
@unindented do you have an ETA for when you think you'll be able to complete this? As far as I know, this is the only thing holding up a non-beta release. (We've got 3.0.0-beta fully working with mobile.twitter.com, which has a lot of tests dealing with ensuring normalizr produces correct data, so I'm confident in that stability of the release.)

unindented:
@paularmstrong I'll get this done this weekend.

Dealing with work and a visa application has kept me very busy this week, sorry.

unindented:
@paularmstrong found 10 minutes of peace to finish this. I've tested it with `npm link`, but would it be possible for you to publish another beta version so that I can test it with the normal installation process?

paularmstrong:
@unindented publised v3.0.0-beta.2

unindented:
@paularmstrong `3.0.0-beta.2` worked perfectly on two of my TypeScript projects. It picked up the typings correctly, and all my tests keep passing after migrating to the new schema namespace.

paularmstrong:
Fantastic. Thanks @unindented. I'll merge and cut a non-beta release! 🎉

unindented:
Thanks for getting rid of `lodash`. It has cut my bundle size significantly. 😄

Issue #193, opened 2016-12-27
Schema is not a constructor


kulakowka:
# Problem

1. I created the application using the [create-react-app](https://github.com/facebookincubator/create-react-app).
1. I added normalizr `yarn add normalizr`. (version: `^3.0.0-alpha.5`)
1. I imported a normalizr in index.js:

```diff
import React from 'react'
import ReactDOM from 'react-dom'
import App from './App'
import './index.css'
+ import { normalize, Schema, arrayOf } from 'normalizr'

+ const article = new Schema('articles')
+ const user = new Schema('users')


ReactDOM.render(
  <App />,
  document.getElementById('root')
)
```
4. In the browser console, I see the error: 

![2016-12-27 19 57 22](https://cloud.githubusercontent.com/assets/557190/21504302/bb1b7fa4-cc6e-11e6-8740-a608d307013d.png)

This is because the documentation is outdated? Which version is stable?

PR #194, opened 2016-12-27
Merge v3.0.0-beta to Master

### LINKED ISSUE #54: Should empty arrays be preserved on their entity?
{ id: 1, name: 'test', tags: [] } normalizes as { id: 1, name: 'test' }

Would it make sense to preserve the empty array or at least allow the option to do so?

### LINKED ISSUE #68: Dealing with collections keyed by ID without an ID in the entity itself
Is there any way I can deal with data in the following format?

```
{
  1: {
    name: 'test user'
  }
}
```

As far as I can tell, I can only work with objects that themselves contain an id i.e.

```
{
  name: 'test user',
  id: 1
}
```

### LINKED ISSUE #70: Keep consistency for relations

### LINKED ISSUE #105: Update & Organize Documentation

### LINKED ISSUE #106: Clean up tests
Currently hard to read through. Should be better organized and easier to maintain.

### LINKED ISSUE #107: Add examples
Examples are currently in external projects. We shouldn't rely on them always being available.

### LINKED ISSUE #126: normalize is take some time if object has more number of keys.
# Problem

normalization is take some time if object has more number of keys.  I tried with array has 100 object and each object has 30 keys and mapping with two nested schema. It takes around ~25ms. I tried using schema based iteration it takes ~7ms. I don't know, Is any problem using schema based iteration? Please refer the link below.

https://github.com/vimalceg/normalizr

### LINKED ISSUE #165: Django rest framework and normalizr
For a following schema (expressed in django models)

    class A:
      b = foreigneKey(B)
      
    
    class B:
      pass

Django rest framework serializes A instance as

    { 
      'id': 1,
      'b': 3
    }

normalizr seems to want a format of

    {
      'id': 1,
      'b': {
        'id': 3
      }
    }

I could change DRF 's default PrimaryKeyRelatedField's serialization, but I 'm wondering if there's another way?  
Maybe there's a way to tell normalizr that 'b' is itself an id attribute?

### LINKED ISSUE #168: http://rackt.github.io/redux goes tp 404
link on the readme http://rackt.github.io/redux goes to 404

### LINKED ISSUE #185: Request: Add a changelog
I'm attempting to debug in a breaking change introduced in 2.3

My first step is usually to run through the changelog, but to my surprise there is none to be found (At least to my eyes, but I have been known to miss things).

Also, want to say thanks for such an awesome project!

### LINKED ISSUE #186: results for arrayOf returns array instead of map as of version 2.3.0
# Problem

Upgrading to 2.3.0 introduces a breaking change to the way `results` are returned.

It looks like results are now normalized when both values are the same. https://github.com/paularmstrong/normalizr/blob/master/src/index.js#L116

This is a breaking change and should result in a major bump and a warning to those who expect a keyed map in return regardless of the values of that map.

**Input**

```js
const challenge = new Schema('challenge', { idAttribute: 'dashedName' });
const block = new Schema('block', { idAttribute: 'dashedName' });
const superBlock = new Schema('superBlock', { idAttribute: 'dashedName' });

block.define({
  challenges: arrayOf(challenge)
});

superBlock.define({
  blocks: arrayOf(block)
});

const mapSchema = valuesOf(superBlock);
const map = {
  'getting-started': {
    dashedName: 'getting-started',
    blocks: [
      {
        dashedName: 'some-block',
        challenges: [{
          dashedName: 'some-challenge',
          ...
        },
        ...
       ],
        ...
      },
      ...
    ],
    ...
  },
   ...
};
normalize(map, mapSchema);
```

**Output**
Here is what has been the output up to 2.2.1
```js
 {
   result: {
  'getting-started': 'getting-started',
  'front-end-development-certification': 'front-end-development-certification',
  'data-visualization-certification': 'data-visualization-certification',
  'back-end-development-certification': 'back-end-development-certification',
  'video-challenges': 'video-challenges',
  'full-stack-development-certification': 'full-stack-development-certification',
  'coding-interview-preparation': 'coding-interview-preparation'
},
   entities: { ... }
}
```

Here is what I actually as of 2.3.0

```js
{
  result: [ [ 
  'getting-started',
  'front-end-development-certification',
  'data-visualization-certification',
  'back-end-development-certification',
  'video-challenges',
  'full-stack-development-certification',
  'coding-interview-preparation'
  ],
  entities: { ...}
}
```

### LINKED ISSUE #187: v3 Docs
# Problem

Documentation in <= 2.x has been a bit difficult to follow, with lots of really long examples in a single README.

# Solution

For v3, create a `/docs` folder and separate out meaningful documentation by section/page, with shorter, easier to follow examples.

### LINKED ISSUE #188: Add In-Repo Examples
# Problem

Examples are currently somewhere untracked by this project. They have moved or been lost in the past.

# Solution

Add an `/examples` folder with clear and concise usage patterns.

paularmstrong:
# Problem

There may be some confusion with users coming from fresh installs of Normalizr. `npm install normalizr` and `yarn add normalizr` automatically grab the most recent release, no matter if it's *alpha*, *beta*, or other status.

See #193.

# Solution

Merge branch *normalizr3* into *master*. I've just published `v3.0.0-beta`, which I've got fully working and conforming against (what I believe is) our largest user, mobile.twitter.com.

# Motivation for v3.0.0

* The `assignEntity` function in normalizr@>=2 is the cause of some performance issues.
* Function signatures within normalizr have become bloated and prevent users from doing things like referencing entity parents while trying to infer schema and generate IDs.
* Maintaining the code is becoming a bit difficult
* Maintaining the tests is a nightmare.
* v3.0.0:
  - is about 35% the byte-size (uglified, gzipped)
  - runs at ~2x the speed
  - has a clear schema structure that is easy to extend
  - has 100% test coverage
  - has clear and easy to follow examples in the repo

# TODO

- [X] Add & Update Tests
- [X] Ensure CI is passing (lint, tests, flow)
- [X] Update relevant documentation
- [x] Update references to `normalizr3` branch to `master`.  Will do immediately before merging.

# Issues

* Closes #54 
* Closes #165 
* Closes #188 
* Closes #68 
* Closes #106
* Closes #105
* Closes #107 
* Closes #126
* Closes #186
* Closes #168 
* Closes #187
* Closes #185 
* Closes #70


examples/redux/src/redux/modules/commits.js
```diff
@@ -0,0 +1,44 @@
...
+  const state = getState();
+  const owner = Repo.selectOwner(state);
+  const repo = Repo.selectRepo(state);
+  return api.repos.getCommits({
+    owner,
+    repo
+  }).then((response) => {
+    const data = normalize(response, [ schema.commit ]);
+    dispatch(Users.addUsers(data.entities.users));
+    dispatch(addCommits(data.entities.commits));
```
unindented: Triggering multiple actions here may be considered bad practice. I think the general consensus is that it's better to trigger a single action that maps to some user interaction (e.g. `LOAD_COMMITS`) in order to avoid inconsistent intermediate states and unnecessary re-renders.

examples/redux/src/redux/modules/issues.js
```diff
@@ -0,0 +1,48 @@
...
+  const repo = Repo.selectRepo(state);
+  return api.issues.getForRepo({
+    owner,
+    repo
+  }).then((response) => {
+    const data = normalize(response, [ schema.issue ]);
+    dispatch(Labels.addLabels(data.entities.labels));
+    dispatch(Milestones.addMilestones(data.entities.milestones));
+    dispatch(Users.addUsers(data.entities.users));
+    dispatch(addIssues(data.entities.issues));
```
unindented: Same here.

examples/redux/src/redux/modules/pull-requests.js
```diff
@@ -0,0 +1,48 @@
...
+  const repo = Repo.selectRepo(state);
+  return api.pullRequests.getAll({
+    owner,
+    repo
+  }).then((response) => {
+    const data = normalize(response, [ schema.pullRequest ]);
+    dispatch(Labels.addLabels(data.entities.labels));
+    dispatch(Milestones.addMilestones(data.entities.milestones));
+    dispatch(Users.addUsers(data.entities.users));
+    dispatch(addPullRequests(data.entities.pullRequests));
```
unindented: Same here.

examples/redux/src/redux/modules/commits.js
paularmstrong: I agree. We use a batching middleware in mobile.twitter.com, but I was hoping to keep the normalizr example clear without using too much redux middleware. I cringed a little adding the thunk, even. Also, the reason to not dispatch multiple is typically React-related, since you don't want to trigger multiple render cycles. Since this doesn't include react and it's just an example, maybe a comment about best practices inline would be better?

paularmstrong: On second thought, a single `ADD_ENTITIES` action that each reducer is able to handle should work... will update.

examples/redux/src/redux/modules/pull-requests.js
paularmstrong: ✔︎ Done

examples/redux/src/redux/modules/commits.js
paularmstrong: ✔︎ Done

examples/redux/src/redux/modules/issues.js
paularmstrong: ✔︎ Done

unindented:
👍

paularmstrong:
Thanks for the review, @unindented!

PR #205, opened 2017-01-04
fix(index.d.ts): SchemaFunction return updated to any

### LINKED ISSUE #204: index.d.ts - processStrategy methodreturns a strign instead of any
# Problem

According to the documentation the `processStrategy` method should return an object.

processStrategy(value, parent, key): Strategy to use when pre-processing the entity. Use this method to add extra data, defaults, and/or completely change the entity before normalization is complete. Defaults to returning a shallow copy of the input entity.

But in the [index.d.ts](https://github.com/paularmstrong/normalizr/blob/master/index.d.ts#L2) the `processStartegy` is defined as a `SchemaFunction` which is a function that returns string.

Also in the [relationship example](https://github.com/paularmstrong/normalizr/blob/c85dfa0d7924fe2fbcbf526ffd5cbd0a3819f672/examples/relationships/schema.js#L32), `processStrategy` returns an object.

When I try to use normalizr in Typescript, the `processStrategy` is expecting to return a string instead of a object.

**Input**

```json
{
    "data": {
       "number": 1337,
       "components": [
           {
               "identifier": "smthg-smthg" ,
               "productCards": [
                  {
                        "identifier": "anthr-smthg-smthg"
                  }
               ]
           }
      ] 
   }
}
```

```Typescript
const components = new schema.Entity("components", {
           productCards: new schema.Array(productCards)
       }, {
               idAttribute: 'identifier',
               processStrategy: (value, parent, key) => {
                   return Object.assign({}, value, { dataId: parent.number}); 
               }
       });
```

**Output**

```js
{
    //...
   component: [ 
       {
             dataId: 1337,
            productCards: [
                 //...
            ]
       }
   ],
  productCards: [
   // ...
  ]
}
```

Here's what I *actually* see when I run the above:

Can't transpile the Typescript due to error in `processStrategy` return type, which expects a string.

Could the index.d.ts be modified to represent the library usage?

afnpires:
Fixes #204.

# Problem

According to the documentation the processStrategy method should return an object.

processStrategy(value, parent, key): Strategy to use when pre-processing the entity. Use this method to add extra data, defaults, and/or completely change the entity before normalization is complete. Defaults to returning a shallow copy of the input entity.

But in the index.d.ts the processStartegy is defined as a SchemaFunction which is a function that returns string.

Also in the relationship example, processStrategy returns an object.

When I try to use normalizr in Typescript, the processStrategy is expecting to return a string instead of a object.

# Solution

Update the index.d.ts to change the SchemaFunction return from sting to any.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,5 @@
 declare namespace schema {
-  export type SchemaFunction = (value: any, parent: any, key: string) => string
+  export type SchemaFunction = (value: any, parent: any, key: string) => any
   export type MergeFunction = (entityA: any, entityB: any) => any
 
   export class Array {
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,7 @@
 declare namespace schema {
-  export type SchemaFunction = (value: any, parent: any, key: string) => any
-  export type MergeFunction = (entityA: any, entityB: any) => any
+  export type StrategyFunction = (value: any, parent: any, key: string) => any;
+  export type SchemaFunction = (value: any, parent: any, key: string) => string;
+  export type MergeFunction = (entityA: any, entityB: any) => any;
 
   export class Array {
     constructor(definition: Schema, schemaAttribute?: string | SchemaFunction)
@@ -10,7 +11,7 @@ declare namespace schema {
   export interface EntityOptions {
     idAttribute?: string | SchemaFunction
     mergeStrategy?: MergeFunction
-    processStrategy?: SchemaFunction
+    processStrategy?: StrategyFunction
   }
 
   export class Entity {
--- a/src/__tests__/typescript/entity.ts
+++ b/src/__tests__/typescript/entity.ts
@@ -1,9 +1,21 @@
 import { normalize, schema } from '../../../index';
 
-const data = {/*...*/};
+type User = {
+  id_str: string,
+  name: string
+};
+
+type Tweet = {
+  id_str: string,
+  url: string,
+  user: User
+};
+
+
+const data = { /*...*/ };
 const user = new schema.Entity('users', {}, { idAttribute: 'id_str' });
 const tweet = new schema.Entity('tweets', { user: user }, {
-    idAttribute: 'id_str',
+  idAttribute: 'id_str',
     // Apply everything from entityB over entityA, except for "favorites"
     mergeStrategy: (entityA, entityB) => ({
       ...entityA,
@@ -11,7 +23,7 @@ const tweet = new schema.Entity('tweets', { user: user }, {
       favorites: entityA.favorites
     }),
     // Remove the URL field from the entity
-    processStrategy: (entity) => {
+    processStrategy: (entity: Tweet, parent, key) => {
       const {url, ...entityWithoutUrl} = entity;
       return entityWithoutUrl;
     }


index.d.ts
```diff
@@ -1,5 +1,5 @@
 declare namespace schema {
-  export type SchemaFunction = (value: any, parent: any, key: string) => string
+  export type SchemaFunction = (value: any, parent: any, key: string) => any
```
paularmstrong: We should have a separate function type for `processStrategy`. `SchemaFunction` is used for `idAttribute`, which should *only* return `string.

paularmstrong:
Any idea why `src/__tests__/typescript.test.js` wasn't failing trying to compile `src/__tests__/typescript/entity.ts`? It uses `processStrategy`, but works totally fine.

afnpires:
I (kinda) figured it out why it passes the test.

For some reason in entity.ts when processStrategy receives an anonymous function `(entity: Tweet, parent, key) => { /*...*/}` it doesn't throw any error. As soon I give a type to one parameter (Tweet type as the first parameter in the example below) it throws an error.

```js
import { normalize, schema } from '../../../index';

type User = {
  id_str: string,
  name: string
};

type Tweet = {
  id_str: string,
  url: string,
  user: User
};

const data = { /*...*/ };
const user = new schema.Entity('users', {}, { idAttribute: 'id_str' });
const tweet = new schema.Entity('tweets', { user: user }, {
  idAttribute: 'id_str',
    // Apply everything from entityB over entityA, except for "favorites"
    mergeStrategy: (entityA, entityB) => ({
      ...entityA,
      ...entityB,
      favorites: entityA.favorites
    }),
    // Remove the URL field from the entity
    processStrategy: (entity: Tweet, parent, key) => {
      const {url, ...entityWithoutUrl} = entity;
      return entityWithoutUrl;
    }
});

const normalizedData = normalize(data, tweet);
```
```
 AssertionError: Semantic: C:/dev/normalizr/src/__tests__/typescript/entity.ts (17,59): Argument of type '{ idAttribute: string; mergeStrategy: (entityA: any, entityB: any) => any; proc
essStrategy: (enti...' is not assignable to parameter of type 'EntityOptions'.
  Types of property 'processStrategy' are incompatible.
    Type '(entity: Tweet, parent: any, key: string) => { id_str: string; user: User; }' is not assignable to type 'SchemaFunction'.
      Type '{ id_str: string; user: User; }' is not assignable to type 'string'.
    at C:\dev\normalizr\node_modules\typescript-definition-tester\dist\index.js:17:15
    at Array.forEach (native)
    at handleDiagnostics (C:\dev\normalizr\node_modules\typescript-definition-tester\dist\index.js:14:17)
    at compile (C:\dev\normalizr\node_modules\typescript-definition-tester\dist\index.js:27:5)
    at C:\dev\normalizr\node_modules\typescript-definition-tester\dist\index.js:55:13
    at next (C:\dev\normalizr\node_modules\typescript-definition-tester\dist\index.js:74:24)
    at C:\dev\normalizr\node_modules\typescript-definition-tester\dist\index.js:87:21
    at C:\dev\normalizr\node_modules\graceful-fs\polyfills.js:287:18
    at FSReqWrap.oncomplete (fs.js:82:15)
```

As soon I return a string inside the method the test yields no error.

```js
import { normalize, schema } from '../../../index';

type User = {
  id_str: string,
  name: string
};

type Tweet = {
  id_str: string,
  url: string,
  user: User
};


const data = { /*...*/ };
const user = new schema.Entity('users', {}, { idAttribute: 'id_str' });
const tweet = new schema.Entity('tweets', { user: user }, {
  idAttribute: 'id_str',
    // Apply everything from entityB over entityA, except for "favorites"
    mergeStrategy: (entityA, entityB) => ({
      ...entityA,
      ...entityB,
      favorites: entityA.favorites
    }),
    // Remove the URL field from the entity
    processStrategy: (entity: Tweet, parent, key) => {
      const {url, ...entityWithoutUrl} = entity;
      return "smthg-smthg";
    }
});

const normalizedData = normalize(data, tweet);
```

afnpires:
Updated the PR to include a modified test for the class and a new type for provideStrategy function.

paularmstrong:
Published v3.0.2 with this fix.

PR #213, opened 2017-01-10
replace file name from 'index' to 'normalizr' in dist folder to keep …

Amorites:
…consistency with older version

# Problem

It's irregular to name your dist files as 'index.js'.


paularmstrong:
Why is this a problem? You shouldn't be pointing directly to that file (in Node.js) and the package.json points to the correct file anyway.

This also breaks usage because now the package.json entry for `"main"` is incorrect.

PR #217, opened 2017-01-12
Add support for custom idAttribute in denormalization

CrOrc:
# Problem

Current denormalization doesn't support custom id attribute

# Solution

Since Entity scheme has getId function we can call it to get entity id.

# Limitations

Currently denormalization should work fine only for string id and id function with one argument.

# TODO

- [ ] Add & update tests

### DIFF
--- a/src/schemas/Polymorphic.js
+++ b/src/schemas/Polymorphic.js
@@ -45,6 +45,11 @@ export default class PolymorphicSchema {
       return value;
     }
     const schema = this.isSingleSchema ? this.schema : this.schema[value.schema];
-    return unvisit(value.id || value, schema, entities);
+    let idOrValue;
+    if (typeof schema.getId === 'function') {
+      idOrValue = schema.getId(value) || value;
+    } else { idOrValue = value.id || value; }
+
+    return unvisit(idOrValue, schema, entities);
   }
 }
--- a/src/schemas/__tests__/Entity.test.js
+++ b/src/schemas/__tests__/Entity.test.js
@@ -147,4 +147,13 @@ describe(`${schema.Entity.name} denormalization`, () => {
 
     expect(denormalize(1, menuSchema, entities)).toMatchSnapshot();
   });
+  it('can denormalize previously normalized data', () => {
+    const mySchema = new schema.Entity('cats', {}, { idAttribute: 'meow' });
+    const cat = { 'meow': 'Meow', hair: 'none' };
+    const normalizedCat = normalize(cat, mySchema);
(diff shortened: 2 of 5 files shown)


src/schemas/Polymorphic.js
```diff
@@ -45,6 +45,11 @@ export default class PolymorphicSchema {
       return value;
     }
     const schema = this.isSingleSchema ? this.schema : this.schema[value.schema];
-    return unvisit(value.id || value, schema, entities);
+    let idOrValue;
+    if (typeof schema.getId === 'function') {
```
paularmstrong: In a polymorphic transform, the `value` will always either be of the form: `{ id, schema }` or just the `id` value. 

Where are you coming across a different case? Can you include a test that proves this is needed?

src/schemas/__tests__/Entity.test.js
```diff
@@ -147,4 +147,13 @@ describe(`${schema.Entity.name} denormalization`, () => {
 
     expect(denormalize(1, menuSchema, entities)).toMatchSnapshot();
   });
+  it('can denormalize previously normalized data', () => {
```
paularmstrong: Entity doesn't use the `Polymorphic` schema denormalization. This test will pass regardless of the change you made. Looks like an oversight on my part to not have included an Entity test with a custom schemaAttribute, but that's mostly because it doesn't matter—the lookup for denormalize goes by key and doesn't dig into the entity to find the ID.

src/schemas/Polymorphic.js
CrOrc: Normalization doesn't add id to entity. I'm coming from roundtrip point of view.
```
  it('can denormalize previously normalized data', () => {
    const mySchema = new schema.Entity('cats', {}, { idAttribute: 'meow' });
    const cat = { 'meow': 'Meow', hair: 'none' };
    const normalizedCat = normalize(cat, mySchema);
    expect(normalizedCat).toMatchSnapshot();
    const denormalizedCat = denormalize(cat, mySchema, normalizedCat.entities);
    expect(denormalizedCat).toEqual(cat);
    expect(denormalizedCat).toMatchSnapshot();
  });
```

paularmstrong: See my other comment. This still works regardless of the other change.

src/schemas/__tests__/Entity.test.js
CrOrc: Sorry about misinterpretation. Figured it out when testing Union schema.

I'm closing this pull request.

Issue #216, opened 2017-01-11
Problem with denormalize method on recursive linked entities


maxsbelt:
# Problem

Problem occurs when I tried to use `denormalize` method when `entities` object contains recursive keys.

**Input**

Example below describes the problem.

```js
const userEntity = new schema.Entity('users')
const reportEntity = new schema.Entity('reports')

userEntity.define({
  reports: [reportEntity]
})
reportEntity.define({
  user: userEntity
})

const entities = {
  reports: {
    1: {
      title: 'Report 1',
      user: 1
    },
    2: {
      title: 'Report 2',
      user: 2
    }
  },
  users: {
    1: {
      username: 'first',
      reports: [1, 2]
    },
    2: {
      username: 'second',
      reports: []
    }
  }
}

const item = denormalize(entities.reports[1], reportEntity, entities)
```

**Output**

Here's what I expect to see when I run the above:

```js
{
  title: 'Report 1',
  user: {
    username: 'first'
    reports: [
      {
        title: 'Report 1',
        user: 1 // or even without this key
      },
      {
        title: 'Report 2',
        user: 2 // or even without this key
      }
    ]
  }
}
```

Here's what I *actually* see when I run the above:

```js
RangeError: Maximum call stack size exceeded
```

paularmstrong:
O_O You have actual recursive data? If you want your reports to give `user: <userId>`, then don't add the schema definition for `user: userEntity`. Wouldn't that solve the problem?

maxsbelt:
Yes, I works with actual recursive data. It comes then 2 resources return nested entities with cross referenced items:

```js
// api/tasks/1/show
{
  id: 1,
  title: 'Task 1',
  comments: [
    {
      id: 1,
      comment: 'Comment 1 to task 1'
    }
  ]
}
```

and

```js
// api/comments/1/show
{
  id: 1,
  comment: 'Comment 1 to task 1',
  task: {
    id: 1,
    title: 'Task 1'
  }
}
```

Actually, I already solved my problem by creating a class that extends `schema.Entity` and redefined `denormalize` method.

I think you need to update documentation of `denormalize` method to specify that denormalizing of recursive schemas is not supported.

paularmstrong:
@maxsbelt PRs accepted

jpstrikesback:
Also discovered this :) @maxsbelt have a repo we could see? (or a PR?)

paularmstrong:
[...]

PR #220, opened 2017-01-13
Added recursive dependencies support for denormalize method

maxsbelt:
# Problem

`denormalize` method doesn't support [recursive linked entities](https://github.com/paularmstrong/normalizr/issues/216).

# Solution

`__cache` key is added to `entities` object parameter for keeping already visited entities. The same approach is already used in [denormalizr](https://github.com/gpbl/denormalizr/commit/81df0765af65dbf19c85773e1f36978debf4f29e) package.

### DIFF
--- a/src/__tests__/__snapshots__/index.test.js.snap
+++ b/src/__tests__/__snapshots__/index.test.js.snap
@@ -33,6 +33,21 @@ Object {
 }
 `;
 
+exports[`denormalize denormalizes recursive dependencies 1`] = `
+Object {
+  "title": "Weekly report",
+  "user": Object {
+    "reports": Array [
+      Object {
+        "title": "Weekly report",
+        "user": Object {},
+      },
+    ],
+    "role": "manager",
+  },
+}
+`;
+
 exports[`normalize can use fully custom entity classes 1`] = `
 Object {
   "entities": Object {
--- a/src/__tests__/__snapshots__/index.test.js.snap
+++ b/src/__tests__/__snapshots__/index.test.js.snap
@@ -35,21 +35,34 @@ Object {
 
 exports[`denormalize denormalizes recursive dependencies 1`] = `
 Object {
+  "id": 1,
   "title": "Weekly report",
   "user": Object {
+    "id": 1,
     "reports": Array [
+      1,
+    ],
+    "role": "manager",
+  },
+}
+`;
+
+exports[`denormalize denormalizes recursive dependencies with custom idAttribute 1`] = `
+Object {
+  "title": "Weekly report",
+  "user": Object {
+    "id": 1,
+    "reports": Array [
+      1,
       Object {
-        "title": "Weekly report",
-        "user": Object {
-          "reports": Array [
-            1,
-          ],
-          "role": "manager",
-        },
+        "title": "Monthly report",
+        "user": 1,
+        "uuid": 2,
       },
     ],
     "role": "manager",
   },
+  "uuid": 1,
 }
 `;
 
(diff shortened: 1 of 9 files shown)


src/__tests__/__snapshots__/index.test.js.snap
```diff
@@ -33,6 +33,21 @@ Object {
...
 `;
 
+exports[`denormalize denormalizes recursive dependencies 1`] = `
+Object {
+  "title": "Weekly report",
+  "user": Object {
+    "reports": Array [
+      Object {
+        "title": "Weekly report",
+        "user": Object {},
```
paularmstrong: I would expect this to be the user's ID, not an empty object. I don't think this is actually working the way you want.

src/schemas/Object.js
```diff
@@ -17,7 +17,21 @@ export const denormalize = (schema, input, unvisit, entities) => {
...
+      if (Array.isArray(object[key])) {
+        object[key] = unvisit(object[key], localSchema, entities);
+      } else {
+        const skey = localSchema.key;
+        if (!entities.__cache[skey]) {
+          entities.__cache[skey] = {};
+        }
+
+        if (!entities.__cache[skey][object[key]]) {
+          entities.__cache[skey][object[key]] = {};
```
paularmstrong: Line is unnecessary since you're definiting a value for it on the next line.

src/schemas/Object.js
```diff
@@ -17,7 +17,21 @@ export const denormalize = (schema, input, unvisit, entities) => {
   Object.keys(schema).forEach((key) => {
     const localSchema = schema[key];
     if (object[key]) {
-      object[key] = unvisit(object[key], localSchema, entities);
+      if (Array.isArray(object[key])) {
+        object[key] = unvisit(object[key], localSchema, entities);
+      } else {
+        const skey = localSchema.key;
```
paularmstrong: What does `skey` stand for? No need to shorten variable names and sacrifice clarity.

src/index.js
```diff
@@ -72,5 +72,6 @@ export const denormalize = (input, schema, entities) => {
     return input;
   }
 
+  entities = { ...entities, __cache: {} };
```
paularmstrong: I don't like adding a reserved key here. There's no reason someone might have a `__cache` entity. (I know, it's not the best name for an entity, but I'd hate to ruin someone's day if they tried to use it).

I'd rather see a separate `visitedEntities` cache object passed around than using reserved keys.

src/schemas/Object.js
maxsbelt: Actually this line block the recursion, but I agree with you that `{}` is unexpected value. Fixed this by passing object from entities to it.

src/__tests__/index.test.js
```diff
@@ -206,4 +206,40 @@ describe('denormalize', () => {
...
+      user: user
+    });
+
+    const entities = {
+      reports: {
+        1: {
+          title: 'Weekly report',
+          user: 1
+        },
+        2: {
```
paularmstrong: Why do you have all of this extra setup data (report 2, user 2). Can delete since it's not being used.

src/__tests__/__snapshots__/index.test.js.snap
```diff
@@ -33,6 +33,26 @@ Object {
 }
 `;
 
+exports[`denormalize denormalizes recursive dependencies 1`] = `
+Object {
```
paularmstrong: My expected output would be:

```js
{
  title: "Weekly report",
  user: {
    role: 'manager',
    reports: [ 1 ]
  }
}
```

This is going too many levels deep and isn't properly keeping track of the fact that we've already come across report 1.

maxsbelt: It's reasonable. I changed denormalization logic to confirm described expectations. Please, review last [commit](https://github.com/paularmstrong/normalizr/pull/220/commits/064ae64d4ca391a333204f4ddb3e8ab5a0ce3a71).

Issue #225, opened 2017-01-24
denormalize crashes when a nested entity is originally null


ChenRoth:
# Problem

`denormalize` crashes when a nested entity is originally `null`

**Input**

```js
const child = new schema.Entity('children')

const user = new schema.Entity('users', {
  children: [child],
})

const mySchema = { users: [user] }

const entities = {
  users: {
    1: { id: 1, children: [{ c: 1 }, { c: 2 }] },
    2: { id: 2, children: null }
  }
};
const denormalizedData = denormalize({ users: [1, 2] }, mySchema, entities);
```

**Output**

Here's what I expect to see when I run the above:

```js
{ users: 
   [ { id: 1, children: [ { c: 1 }, { c: 2 } ] },
     { id: 2, children: null } ] }
```

Here's what I *actually* see when I run the above:

```
.../normalizr/dist/src/schemas/Array.js:51
  return input.map(function (entityOrId) {
              ^

TypeError: Cannot read property 'map' of null
```

CoinGame:
I'm running into this problem right now. Is there a suggested workaround other than ensuring all null entity values are empty arrays first?

ChenRoth:
@CoinGame not that i know of.. i replaced the `null` with an empty list too.
If I ever get the time soon, I'll try to submit a fix

paularmstrong:
This should be a simple fix in the ArraySchema denormalize loop. Definitely open to PRs, thought I won't be back until some time next week.

CoinGame:
I'm still a  bit of a javascript noob, so I'm not sure if this is a safe change. I modified Array.js to check that the value was an array before trying to map. It's not breaking for me anymore. 

```javascript
  return typeof(input) === Array ? input.map(function (entityOrId) {
    return unvisit(entityOrId, schema, entities);
  }) : null;
```
https://github.com/paularmstrong/normalizr/blob/master/src/schemas/Array.js#L24-L27

paularmstrong:
I had to overhaul a few things and this was fixed as a result. Will be out in the next release (hopefully this week)

CoinGame:
Fantastic. Thank you for the quick update.

PR #228, opened 2017-01-25
Support denormalizing from immutable entities

### LINKED ISSUE #222: denormalize with immutable js
Will support be added for denormalizing immutable js objects?

jcarbo:
Fixes #222 

# Problem

First I want to say that I've been using and following this repo for quite a while and it's great to see `denormalize` finally land! 

It seems that most places I've seen that suggest using `normalizr` also suggest using an immutable store for entities. However, in doing so you lose the ability to use `denormalize`. Adding `immutable` as a dependency would add complexity and bloat for users who are not using immutable data.

# Solution

Update `denormalize` to handle both mutable and immutable entities without actually requiring `immutable` as a dependency.

Support is handled by using simple `getIn`, `setIn`, `hasIn` methods that defer to the immutable implementations if the object is immutable otherwise implement the same behavior on plain javascript objects/arrays. `immutable-js` is required as a dev dependency since it's needed for specs. I just updated the existing specs to also test with immutable entities.

The best part is that it only adds ~70 bytes (compressed) and ~40 bytes (minified + compressed) and doesn't add any external dependencies.

# TODO

- [x] Add & update tests

# Size Diff

## Before

```
Package:      dist/normalizr.js
Bundle Size:  18.06 KB
Compressed:   3.93 KB

Package:      dist/normalizr.min.js
Bundle Size:  8.76 KB
Compressed:   2.48 KB
```
## After

```
Package:      dist/normalizr.js
Bundle Size:  19.31 KB
Compressed:   4.09 KB

Package:      dist/normalizr.min.js
Bundle Size:  9.25 KB
Compressed:   2.61 KB
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,6 +3,7 @@ import UnionSchema from './schemas/Union';
 import ValuesSchema from './schemas/Values';
 import ArraySchema, * as ArrayUtils from './schemas/Array';
 import ObjectSchema, * as ObjectUtils from './schemas/Object';
+import * as ImmutableUtils from './schemas/ImmutableUtils';
 
 const visit = (value, parent, key, schema, addEntity) => {
   if (typeof value !== 'object' || !value) {
@@ -67,7 +68,9 @@ const getEntities = (entities, visitedEntities) => (schema, entityOrId) => {
     visitedEntities[schemaKey] = {};
   }
 
-  const entity = typeof entityOrId === 'object' ? entityOrId : entities[schemaKey][entityOrId];
+  const entity = typeof entityOrId === 'object' ?
+    entityOrId :
+    ImmutableUtils.getIn(entities, [ schemaKey, entityOrId ]);
   const id = schema.getId(entity);
   if (visitedEntities[schemaKey][id]) {
     return id;
--- a/src/index.js
+++ b/src/index.js
@@ -62,25 +62,23 @@ const unvisit = (input, schema, getDenormalizedEntity) => {
   return schema.denormalize(input, unvisit, getDenormalizedEntity);
 };
 
-const getEntity = (entityOrId, schemaKey, entities) => {
+const getEntity = (entityOrId, schemaKey, entities, isImmutable) => {
   if (typeof entityOrId === 'object') {
     return entityOrId;
   }
 
-  if (ImmutableUtils.isImmutable(entities)) {
-    return entities.getIn([ schemaKey, entityOrId.toString() ]);
-  }
-
-  return entities[schemaKey][entityOrId];
+  return isImmutable ?
+    entities.getIn([ schemaKey, entityOrId.toString() ]) :
+    entities[schemaKey][entityOrId];
 };
 
-const getEntities = (entities, visitedEntities) => (schema, entityOrId) => {
+const getEntities = (entities, visitedEntities, isImmutable) => (schema, entityOrId) => {
   const schemaKey = schema.key;
   if (!visitedEntities[schemaKey]) {
     visitedEntities[schemaKey] = {};
   }
 
-  const entity = getEntity(entityOrId, schemaKey, entities);
+  const entity = getEntity(entityOrId, schemaKey, entities, isImmutable);
   const id = schema.getId(entity);
   if (visitedEntities[schemaKey][id]) {
     return id;
@@ -95,6 +93,7 @@ export const denormalize = (input, schema, entities) => {
     return input;
   }
 
-  const getDenormalizedEntity = getEntities(entities, {});
+  const isImmutable = ImmutableUtils.isImmutable(entities);
+  const getDenormalizedEntity = getEntities(entities, {}, isImmutable);
   return unvisit(input, schema, getDenormalizedEntity);
 };
--- a/src/index.js
+++ b/src/index.js
@@ -62,15 +62,25 @@ const unvisit = (input, schema, getDenormalizedEntity) => {
   return schema.denormalize(input, unvisit, getDenormalizedEntity);
 };
 
+const getEntity = (entityOrId, schemaKey, entities) => {
+  if (typeof entityOrId === 'object') {
+    return entityOrId;
+  }
+
+  if (ImmutableUtils.isImmutable(entities)) {
+    return entities.getIn([ schemaKey, entityOrId.toString() ]);
+  }
+
+  return entities[schemaKey][entityOrId];
+};
+
 const getEntities = (entities, visitedEntities) => (schema, entityOrId) => {
   const schemaKey = schema.key;
   if (!visitedEntities[schemaKey]) {
     visitedEntities[schemaKey] = {};
   }
 
-  const entity = typeof entityOrId === 'object' ?
-    entityOrId :
-    ImmutableUtils.getIn(entities, [ schemaKey, entityOrId ]);
+  const entity = getEntity(entityOrId, schemaKey, entities);
   const id = schema.getId(entity);
   if (visitedEntities[schemaKey][id]) {
     return id;
--- a/src/index.js
+++ b/src/index.js
@@ -59,6 +59,10 @@ const unvisit = (input, schema, getDenormalizedEntity) => {
     return method(schema, input, unvisit, getDenormalizedEntity);
   }
 
+  if (input === undefined || input === null) {
+    return input;
+  }
+
   return schema.denormalize(input, unvisit, getDenormalizedEntity);
 };
 
--- a/src/schemas/Entity.js
+++ b/src/schemas/Entity.js
@@ -1,3 +1,5 @@
+import * as ImmutableUtils from './ImmutableUtils';
+
 export default class EntitySchema {
   constructor(key, definition = {}, options = {}) {
     if (!key || typeof key !== 'string') {
@@ -13,7 +15,9 @@ export default class EntitySchema {
     } = options;
 
     this._key = key;
-    this._getId = typeof idAttribute === 'function' ? idAttribute : (input) => input[idAttribute];
+    this._getId = typeof idAttribute === 'function' ?
+      idAttribute :
+      (input) => ImmutableUtils.getIn(input, [ idAttribute ]);
     this._idAttribute = idAttribute;
     this._mergeStrategy = mergeStrategy;
     this._processStrategy = processStrategy;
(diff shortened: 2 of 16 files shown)


src/schemas/__tests__/__snapshots__/Array.test.js.snap
```diff
@@ -11,6 +11,19 @@ Array [
 ]
 `;
 
+exports[`ArraySchema denormalization Class denormalizes a single entity 2`] = `
+Array [
+  Object {
+    "id": 1,
+    "name": "Milo",
+  },
```
jcarbo: It seems that immutable data structures are converted to plain objects when serialized into snapshots, see: https://github.com/facebook/jest/issues/1622. In this case, the Array should be an actual array since the test uses `[1,2]` as input but each object is actually an `Immutable.Map` (which I manually verified is the case).

paularmstrong:
Nice work @jcarbo! For the most part this looks good. I have a few nit requests, I think, but I'd like to spend more time with it once I'm back (currently traveling till next week)

src/schemas/Entity.js
```diff
@@ -52,12 +54,12 @@ export default class EntitySchema {
   }
 
   denormalize(entityOrId, unvisit, entities) {
-    const entity = typeof entityOrId === 'object' ? entityOrId : entities[this.key][entityOrId];
-    const entityCopy = { ...entity };
+    const entity = typeof entityOrId === 'object' ? entityOrId : getIn(entities, [ this.key, entityOrId ]);
+    let entityCopy = isImmutable(entity) ? entity : { ...entity };
```
paularmstrong: Avoid using `let`, please

src/schemas/Entity.js
```diff
@@ -52,12 +54,12 @@ export default class EntitySchema {
...
     Object.keys(this.schema).forEach((key) => {
-      if (entityCopy.hasOwnProperty(key)) {
+      if (hasIn(entityCopy, [ key ])) {
```
paularmstrong: Can we use the `isImmutable(entity)`  method before doing `getIn`, `hasIn`, and `setIn` to avoid unnecessary overhead?

src/schemas/ImmutableUtils.js
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Check if an object is immutable by checking if it implements the
+ * getIn method.
+ *
+ * @param  {any} object
+ * @return {bool}
+ */
+export function isImmutable(object) {
+  return !!(object && object.getIn);
```
paularmstrong: edge-case: what if someone has a key, `getIn` on their entity? (also true for `setIn` and `hasIn`.

src/schemas/Object.js
```diff
@@ -13,14 +15,14 @@ export const normalize = (schema, input, parent, key, visit, addEntity) => {
 };
 
 export const denormalize = (schema, input, unvisit, entities) => {
-  const object = { ...input };
+  let inputCopy = isImmutable(input) ? input : { ...input };
```
paularmstrong: nit: avoid `let` statements.

src/schemas/Object.js
```diff
@@ -1,3 +1,5 @@
+import { getIn, hasIn, isImmutable, setIn } from './ImmutableUtils';
```
paularmstrong: nit: `import * as ImmutableUtils from './ImmutableUtils';`

jcarbo:
@paularmstrong - thanks for reviewing! I rebased and got the tests back to green.

Also, reading through your comments it seems the your (valid) concerns were around performance/overhead and robustness. The commit I just pushed up takes a slightly different approach and should address your concerns:
- to reduce false positives, make the `isImmutable` check much more specific. This is a little brittle given it's relying on some internals of `Immutable.Map` and `Immutable.Record` but I think the tradeoff is more than worth it to avoid bringing in `immutable` as a dependency. Also the `immutable` library is one of the more stable ones I've seen in the javascript world so it's unlikely we'll be affected.
- There was a lot indirection introduced so that the same `denormalize` functions could handle both immutable and mutable objects. Instead, at the top of the `denormalize` function in the `Entity` and `Object` schemas we first check if the object is immutable and, if so, pass it to a separate `denormalizeImmutable` record.

With these changes, the compressed size increased by another ~90 bytes. Still not that bad imho.

src/index.js
```diff
@@ -61,13 +62,25 @@ const unvisit = (input, schema, getDenormalizedEntity) => {
   return schema.denormalize(input, unvisit, getDenormalizedEntity);
 };
 
+const getEntity = (entityOrId, schemaKey, entities) => {
+  if (typeof entityOrId === 'object') {
+    return entityOrId;
+  }
+
+  if (ImmutableUtils.isImmutable(entities)) {
```
paularmstrong: This could be sent in as a flag, since `entities` is a wrapped argument in the `getEntities` function. This would save on this conditional over every entity iteration. Something like...

```js
const getEntities = (entities, visitedEntities) => {
  const isImmutable = ImmutableUtils.isImmutable(entities);
  return (schema, entityOrId) => {
    // ...
    const entity = getEntity(entityOrId, schemaKey, entities, isImmutable);
    // ...
  };
};
```

paularmstrong: This could be sent in as a flag, since `entities` is a wrapped argument in the `getEntities` function. This would save on this conditional over every entity iteration. Something like...

```js
const getEntities = (entities, visitedEntities) => {
  const isImmutable = ImmutableUtils.isImmutable(entities);
  return (schema, entityOrId) => {
    // ...
    const entity = getEntity(entityOrId, schemaKey, entities, isImmutable);
    // ...
  };
};
```

src/schemas/Entity.js
```diff
@@ -13,7 +15,9 @@ export default class EntitySchema {
     } = options;
 
     this._key = key;
-    this._getId = typeof idAttribute === 'function' ? idAttribute : (input) => input[idAttribute];
+    this._getId = typeof idAttribute === 'function' ?
+      idAttribute :
+      (input) => ImmutableUtils.isImmutable(input) ? input.get(idAttribute) : input[idAttribute];
```
paularmstrong: Can we avoid nested ternaries, please? They get difficult to read.

src/schemas/ImmutableUtils.js
```diff
@@ -0,0 +1,41 @@
...
+ * @param  {Immutable.Map|Immutable.Record} input
+ * @param  {function} unvisit
+ * @param  {function} getDenormalizedEntity
+ * @return {Immutable.Map|Immutable.Record}
+ */
+export function denormalizeImmutable(schema, input, unvisit, getDenormalizedEntity) {
+  return Object.keys(schema).reduce((object, key) => {
+    // Immutable maps cast keys to strings on write so we need to ensure
+    // we're accessing them using string keys.
+    const stringKey = typeof key === 'string' ? key : key.toString();
```
paularmstrong: Would be safe to just force casting it:

```js
const stringKey = `${key}`;
```

jcarbo:
Just pushed those changes 👍

jcarbo:
Friendly nudge on this when you get a chance :)

src/schemas/ImmutableUtils.js
```diff
@@ -0,0 +1,41 @@
...
+ * Check if an object is immutable by checking if it has a key specific
+ * to the immutable library.
+ *
+ * @param  {any} object
+ * @return {bool}
+ */
+export function isImmutable(object) {
+  return !!(object && (
+    object.hasOwnProperty('__ownerID') || // Immutable.Map
+    (object._map && object._map.object.hasOwnProperty('__ownerID')) // Immutable.Record
```
paularmstrong: This condition is never hit by tests. Can you add an case for this?

jcarbo:
Good call on that missing coverage, the conditional was actually broken!

src/index.js
```diff
@@ -58,16 +59,30 @@ const unvisit = (input, schema, getDenormalizedEntity) => {
     return method(schema, input, unvisit, getDenormalizedEntity);
   }
 
+  if (input === undefined || input === null) {
```
jcarbo: Wanted to call your attention to this change added in the last commit. This came up in the test with immutable records because it always instantiates all fields defined in the schema, using the default value if you didn't give it pass it one. E.g.

```js
> Menu = new Immutable.Record({ id: null, food: null })
[Function: Record]
> menuData = { id: 2 }
{ id: 2 }
> nonRecordMenu = menuData
{ id: 2 }
> recordMenu = new Menu(menuData)
Record { "id": 2, "food": null }
```

So while denormalizing the record, it's going to try to unvisit `food: null` which currently throws an unexpected error in `_getId`

This assumes that you're not going to have `undefined` or `null` resolve to anything during the denormalization process. Is that an ok assumption to make?

paularmstrong: Yeah that's a good one and similar to #225. This is good. Thanks!

paularmstrong:
Nice job on this! Looks like we've actually increased the performance since the initial release of normalizr2 and this addition doesn't have any ill effects! We used to be right at about 200%. Now looking at 220%.

```
normalzr2 x 111 ops/sec ±4.77% (70 runs sampled)
normalizr3 x 246 ops/sec ±4.60% (73 runs sampled)
```

src/index.js
jcarbo: Another related thing I just realized: this currently throws that same unexpected error if an entity is not found:

```
const entities = {
  tacos: {
    1: { id: 1, type: 'foo' }
  }
};

> denormalize(3, mySchema, entities)
TypeError: Cannot read property 'id' of undefined
    at EntitySchema._getId (~/normalizr/dist/normalizr.js:226:65)
    at EntitySchema.getId (~/normalizr/dist/normalizr.js:245:19)
    at ~/normalizr/dist/normalizr.js:635:21
    at EntitySchema.denormalize (~/normalizr/dist/normalizr.js:273:20)
    at unvisit (~/normalizr/dist/normalizr.js:612:17)
    at Object.denormalize$1 [as denormalize] (~/normalizr/dist/normalizr.js:651:10)
    at repl:1:3
    at realRunInThisContextScript (vm.js:22:35)
    at sigintHandlersWrap (vm.js:98:12)
    at ContextifyScript.Script.runInThisContext (vm.js:24:12)
```

I kinda feel like that should just return `undefined`. In my apps, I think I'd rather just ignore missing data, especially when denormalizing nested data.

I could also see it being configurable so you can handle how you want (e.g. initialize with empty object, return null,undefined, throw error)
```
denormalize(3, mySchema, entities, { onNotFound: (id, schema) => undefined })
denormalize(3, mySchema, entities, { onNotFound: (id, schema) => { id } })
denormalize(3, mySchema, entities, { onNotFound: (id, schema) => throw new Error(`could not find ${schema.name}` })
```

Thoughts? If you agree I'd probably remove this line and just handle that below if an entity is not found in the store.

paularmstrong: I think there was a PR for this and it was randomly closed... #229. Makes sense to fix. We don't ever run across that case, but I could see it being possible.

paularmstrong: Also, already merged this PR... so maybe a new one?

paularmstrong: Sorry, my brain isn't working today... I'd rather stay away from the extra configuration, as you proposed. I'd expect the application to handle receiving `null` for an entity elsewhere than pushing more into normalizr.

PR #229, opened 2017-01-26
return null instead of empty object

maartenschumacher:
# Problem

When an entity is not found during denormalisation, it is supplanted by an empty object. This makes it impossible to guarantee the shape of an object at runtime.

# Solution

When an entity isn't found, even if it's deeply nested, the result of the whole denormalisation should be null.

### DIFF
--- a/src/schemas/Entity.js
+++ b/src/schemas/Entity.js
@@ -53,14 +53,25 @@ export default class EntitySchema {
 
   denormalize(entityOrId, unvisit, entities) {
     const entity = typeof entityOrId === 'object' ? entityOrId : entities[this.key][entityOrId];
+    if (!entity) {
+      return null;
+    }
+    let failure = false;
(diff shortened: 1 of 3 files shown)


shulcsm:
Why did you close this? I'm having same issue. I want nulls for optional relations.

maartenschumacher:
Sorry, I actually meant to open this PR on a fork, but opened it here by accident.

paularmstrong:
This was fixed in #232, available in v3.2.0

salanki:
Wouldn't it make more sense to return `undefined` instead of null if the object is missing, just as an entry that is not part of a schema would behave?

salanki:
⬆️

shulcsm:
undefined is fine if key is not null but entity is missing from map. If key is present and is null, it should stay null. Not sure what is current behavior it has been a while.

Issue #233, opened 2017-02-10
Denormalize method replace a key with the entity only the first entry


Serenya:
I don't know if it's expected behavior or not, but lets take example from the documentation of denormalize method and extend it a little so that user Entity contains reference to a child Entity:
```javascript
import { denormalize, schema } from 'normalizr';

const mentor = new schema.Entity('mentors');
const user = new schema.Entity('users', {
  mentor: mentor
});

const mySchema = { users: [ user ] }
const entities = { users: { '1': { id: 1, mentor: 3 }, '2': { id: 2, mentor: 3 } , mentors: { '3': { id: 3, name: 'John' } } } };
const denormalizedData = denormalize({ users: [ 1, 2 ] }, mySchema, entities);
```

The output is:
```javascript
{ 
  users: [
    { id: 1, mentor: { id: 3, name: 'John' } },
    { id: 2, mentor: 3 }
  ]
}
```
I was expecting something like this:
```javascript
{ 
  users: [
    { id: 1, mentor: { id: 3, name: 'John' } },
    { id: 2, mentor: { id: 3, name: 'John' } }
  ]
}
```
I've found out that in this file https://github.com/paularmstrong/normalizr/blob/master/src/index.js
under method getEntities, you may find the guilty code:
```javascript
const id = schema.getId(entity);
if (visitedEntities[schemaKey][id]) {
  return id;
}
```
Do I understand the concept of denormalization wrong? Is it expected behavior? What is the purpose of having visitedEntities variable?

paularmstrong:
Yep, looks like a bug that appeared in the newest release. We added some recursive-denormalization handling, but it looks like it's not taking into account items seen on multiple paths.

paularmstrong:
PRs appreciated. I don't have time to look into it at the moment. If no one can take this on, I might have to revert it. CC @maxsbelt

paularmstrong:
v3.2.1 is out and reverts the incorrect behavior.

Serenya:
Well done, thanks!

Issue #223, opened 2017-01-18
getAttributeId instance method


karlguillotte:
I am upgrading normilzr to latest from 2.x. 

I am using the instance method `getAttributeId` in my code base. However, it seems that this method has been removed. Any plan on bringing back that function? Or making the `_getId` public?

Thanks for the update and the library! 
Karl

karlguillotte:
Ah! Sorry it is public already as `getId`. 

Perhaps, it could be added to the documentation. 

Thanks!

paularmstrong:
It's not in the documentation because it's not really meant to be publicly visible outside of normalizr. Why are you using it? I don't understand the context for this.

karlguillotte:
Here is a bit of my use-case! 

I am using normalizr in my Redux store along with reselect. I have a selector that take entities for a given schema and transform into GeoJSON objects so I can display them on a map (using Mapbox GL). Inside my selector I am using `getId` to make sure that my GeoJSON objects use the same ids. 

Hope it helps. Do not hesitate if you have concerns/questions or it not clear.

PR #238, opened 2017-02-11
Use a reference cache to allow recursive references

andyhite:
# Problem

Recursive / circular references throw a stack-level too deep error when denormalized.

# Solution

Cache references to the already denormalized entities and use the reference from the cache if it shows back up in a subsequent pass. The bulk of the work is in `src/schemas/Entity.js` – the rest is just passing the cache object around to make sure it's always available to the denormalize method.

### DIFF
--- a/src/__tests__/__snapshots__/index.test.js.snap
+++ b/src/__tests__/__snapshots__/index.test.js.snap
@@ -33,6 +33,20 @@ Object {
 }
 `;
 
+exports[`denormalize denormalizes recursive dependencies 1`] = `
+Object {
+  "id": 1,
+  "title": "Weekly report",
+  "user": Object {
+    "id": 1,
+    "reports": Array [
+      [Circular],
+    ],
+    "role": "manager",
+  },
+}
+`;
+
 exports[`normalize can use fully custom entity classes 1`] = `
 Object {
   "entities": Object {
--- a/src/__tests__/index.test.js
+++ b/src/__tests__/index.test.js
@@ -207,4 +207,34 @@ describe('denormalize', () => {
     });
     expect(() => denormalize('123', article, entities)).not.toThrow();
   });
+
+  it('denormalizes recursive dependencies', () => {
+    const user = new schema.Entity('users');
+    const report = new schema.Entity('reports');
+
+    user.define({
+      reports: [ report ]
+    });
+    report.define({
+      user: user
+    });
+
+    const entities = {
+      reports: {
+        1: {
+          id: 1,
+          title: 'Weekly report',
+          user: 1
+        }
+      },
+      users: {
+        1: {
+          id: 1,
+          role: 'manager',
+          reports: [ 1 ]
+        }
+      }
+    };
+    expect(denormalize('1', report, entities)).toMatchSnapshot();
+  });
 });
--- a/src/index.js
+++ b/src/index.js
@@ -76,10 +76,23 @@ const getEntity = (entityOrId, schemaKey, entities, isImmutable) => {
     entities[schemaKey][entityOrId];
 };
 
-const getEntities = (entities, isImmutable) => (schema, entityOrId) => {
-  const schemaKey = schema.key;
+const getEntities = (entities, cache, isImmutable) => (schema, entityOrId) => {
+  return (handleCacheHit, handleCacheMiss) => {
+    const schemaKey = schema.key;
+    const entityId = typeof entityOrId === 'object' ? schema.getId(entityOrId) : entityOrId;
 
-  return getEntity(entityOrId, schemaKey, entities, isImmutable);
+    if (!cache[schemaKey]) {
+      cache[schemaKey] = {};
+    }
+
+    if (cache[schemaKey][entityId]) {
+      return handleCacheHit(cache[schemaKey][entityId]);
+    }
+
+    const entity = getEntity(entityOrId, schemaKey, entities, isImmutable);
+
+    return handleCacheMiss(entity, cache);
+  };
 };
 
 export const denormalize = (input, schema, entities) => {
@@ -88,6 +101,6 @@ export const denormalize = (input, schema, entities) => {
   }
 
   const isImmutable = ImmutableUtils.isImmutable(entities);
-  const getDenormalizedEntity = getEntities(entities, isImmutable);
+  const getDenormalizedEntity = getEntities(entities, {}, isImmutable);
   return unvisit(input, schema, getDenormalizedEntity);
 };
--- a/src/index.js
+++ b/src/index.js
@@ -79,19 +79,22 @@ const getEntity = (entityOrId, schemaKey, entities, isImmutable) => {
 const getEntities = (entities, cache, isImmutable) => (schema, entityOrId) => {
   return (handleCacheHit, handleCacheMiss) => {
     const schemaKey = schema.key;
-    const entityId = typeof entityOrId === 'object' ? schema.getId(entityOrId) : entityOrId;
+    const getEntityId = (entity) => typeof entity === 'object' ? schema.getId(entity) : entity;
 
     if (!cache[schemaKey]) {
       cache[schemaKey] = {};
     }
 
-    if (cache[schemaKey][entityId]) {
-      return handleCacheHit(cache[schemaKey][entityId]);
+    if (cache[schemaKey][getEntityId(entityOrId)]) {
+      return handleCacheHit(cache[schemaKey][getEntityId(entityOrId)]);
     }
 
     const entity = getEntity(entityOrId, schemaKey, entities, isImmutable);
 
-    return handleCacheMiss(entity, cache);
+    return handleCacheMiss(entity, (entityToCache) => {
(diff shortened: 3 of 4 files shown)


src/index.js
```diff
@@ -82,12 +82,12 @@ const getEntities = (entities, isImmutable) => (schema, entityOrId) => {
...
 };
 
-export const denormalize = (input, schema, entities) => {
+export const denormalize = (input, schema, entities, cache = {}) => {
   if (!input) {
     return input;
   }
 
   const isImmutable = ImmutableUtils.isImmutable(entities);
   const getDenormalizedEntity = getEntities(entities, isImmutable);
```
paularmstrong: The `cache` object should be passed to `getEntities`. The return function for `getEntities` should do the cache logic instead of `schema.Entity`. That'd reduce the amount of passing `cache` around to all of these other methods and centralize the logic in case of future `Entity` types.

src/index.js
```diff
@@ -82,12 +82,12 @@ const getEntities = (entities, isImmutable) => (schema, entityOrId) => {
   return getEntity(entityOrId, schemaKey, entities, isImmutable);
 };
 
-export const denormalize = (input, schema, entities) => {
+export const denormalize = (input, schema, entities, cache = {}) => {
```
paularmstrong: Why do we need to pass a `cache` object into the top level? Seems unnecessary.

andyhite:
I'd love if I didn't need to deal with circular references as well, but unfortunately it's a reality when your data needs to be modeled that way. In my case, I have a situation like this:

* A company has many users
* A user belongs to an account
* The user's company needs to be accessible at `user.company` because I need to display information about the company the user belongs to.
* The company the user belongs to can be edited if they're an administrator, and new users can be added to the company. Therefore I need the logged-in user's company's users accessible at `user.companies.users`.

I guess one solution that I haven't tried would be to create a different schema for each side of the association - `user` and `employee`, which basically model the same thing but use different schemas and employee doesn't define the belongs-to `company` relationship. That may get difficult to maintain as things grow, but it is an alternative.

Anyway, I'll try and implement the changes you suggested - circular references *would* be nice to support on the denormalize side, since they are fully supported on the normalize side.

paularmstrong:
@andyhite I'm not arguing against this change. I hope you didn't get that impression. We had a solution to only pull one entity reference, but that failed for various reasons (#233), so it had to be reverted. I think that your approach makes a lot of sense and I'd love to get it in!

andyhite:
Alright, I just updated this PR with the following changes:

* It no longer passes a cache object around everywhere
* The `getEntities` return function returns another function that handles cache hits / misses. It takes two callback functions, one if there's a cache hit (so the `Entity.denormalize` knows it can return early and not process the sub-entities) and another if there's a cache miss so that the sub-entities can be processed.

It isn't possible to contain the lookup logic to only the `getEntities` function, since the result of `getEntities` is always going to flow through `Entity.denormalize` and therefore still process the entity (which is where the callstack goes nuts). We need to skip the entity processing if there's a cache-hit.

andyhite:
I'm moving a bit more of the logic into the `getEntities` function so `denormalize` will be ignorant of the caching logic...will update the PR shortly.

andyhite:
Updated.

paularmstrong:
I spent a couple hours working on this, but couldn't get recursive Immutable objects working. Do you have time to make that work as well? I'd hate to ship this without full support. You can see what I tried on top of your changes [here](https://github.com/paularmstrong/normalizr/compare/andyhite-master). cc @jcarbo

andyhite:
@paularmstrong no problem. I should have some time later today to take a crack at it.

andyhite:
@paularmstrong does `denormalize` return a regular object data structure if you denormalize an immutable entity? Judging by the tests and snapshots, it looks like it does...is that right?

paularmstrong:
@andyhite it does not. The snapshots look that way because of jest's snapshot formatting.

src/__tests__/__snapshots__/index.test.js.snap
```diff
@@ -33,6 +33,20 @@ Object {
...
+  "user": Object {
+    "id": 1,
+    "reports": Array [
+      [Circular],
+    ],
+    "role": "manager",
+  },
+}
+`;
+
```
jcarbo: The immutable snapshot should be:
```
exports[`denormalize denormalizes recursive dependencies 2`] = `
Object {
  "id": 1,
  "title": "Weekly report",
  "user": Object {
    "id": 1,
    "reports": Array [
      1,
    ],
    "role": "manager",
  },
}
`;
```

src/__tests__/index.test.js
```diff
@@ -207,4 +207,34 @@ describe('denormalize', () => {
...
+      },
+      users: {
+        1: {
+          id: 1,
+          role: 'manager',
+          reports: [ 1 ]
+        }
+      }
+    };
+    expect(denormalize('1', report, entities)).toMatchSnapshot();
```
jcarbo: Add test for immutable on the next line here:
```
denormalize('1', report, fromJS(entities))).toMatchSnapshot();
```

src/index.js
```diff
@@ -76,10 +76,26 @@ const getEntity = (entityOrId, schemaKey, entities, isImmutable) => {
...
 };
 
-const getEntities = (entities, isImmutable) => (schema, entityOrId) => {
-  const schemaKey = schema.key;
-
-  return getEntity(entityOrId, schemaKey, entities, isImmutable);
+const getEntities = (entities, cache, isImmutable) => (schema, entityOrId) => {
+  return (handleCacheHit, handleCacheMiss) => {
+    const schemaKey = schema.key;
+    const getEntityId = (entity) => typeof entity === 'object' ? schema.getId(entity) : entity;
```
jcarbo: Currently you're creating this function on every iteration. Unless I'm missing something, I think you could just set `entityId` at the top of the function and forego this function at all.

```js
return (handleCacheHit, handleCacheMiss) => {
  const schemaKey = schema.key;
  const entityId = typeof entityOrId === 'object' ? schema.getId(entityOrId) : entityOrId;
  // ...
```

If there is a reason for this method to be called each time, I'd create it b/w line 79/80.

jcarbo:
@paularmstrong - Circular references is something that `immutable-js` does not support (and [does not plan to support](https://github.com/facebook/immutable-js/issues/259)). The way we handled this in the standalone denormalizr library (https://github.com/gpbl/denormalizr/pull/12) was to just keep the ID where a circular reference would appear. So using the example from the tests on this PR:

```js
const entities = {
  reports: {
    1: {
      id: 1,
      title: 'Weekly report',
      user: 1
    }
  },
  users: {
    1: {
      id: 1,
      role: 'manager',
      reports: [ 1 ]
    }
  }
};

denormalize('1', report, fromJS(entities));

// Would yield:
Object {
  "id": 1,
  "title": "Weekly report",
  "user": Object {
    "id": 1,
    "reports": Array [
      1,
    ],
    "role": "manager",
  },
}
`;
```

src/index.js
andyhite: It looks like @paularmstrong handled this in the branch that he was trying to implement circular immutable references, which I'll merge into this branch shortly once we've settled on a solution for that.

andyhite:
@jcarbo I'll get this updated to handle `immutable-js` circular references with an ID instead of an actual circular reference. We'll need to be careful to avoid the issue from #223, but I think it can be done by including the parent and key in the cache key.

paularmstrong:
> Circular references is something that immutable-js does not support (and does not plan to support)

@jcarbo Thanks! I haven't used immutable-js for anything yet and I was starting to think this was maybe the case. Thanks for weighing in on this PR!

@andyhite Thanks for bearing with all the feedback!

andyhite:
@paularmstrong no problem!

I believe I may have found an unrelated bug in the handling of immutable stuff – it looks like `List` types, which are basically immutable Arrays, don't get denormalized at all. If you use `fromJS` to turn a mutable structure into an immutable one, Arrays get turned into Lists. If you create a `Record` with a property that has an array value, then the array value remains an Array inside the immutable `Record`. I discovered this when I switched the test @paularmstrong wrote for the recursive immutable over to what @jcarbo suggested (using the `fromJS` bit). This is something that will probably need to be handled, but I'm not sure if this PR is the right place for it.

paularmstrong:
@andyhite Let's handle any unrelated issues separately. Can you file one so we can track it?

vyorkin:
wow! @andyhite thank you for putting your time into this, whats the status of this PR? I'm already using your branch with `npm link` btw :)

andyhite:
I'm still trying to get the immutable bit working, but haven't had time
this week due to other commitments. I'll hopefully have the immutable bit
working so we can get it merged by the end of this week.

jcarbo:
@andyhite @paularmstrong - I gave this a shot in a slightly different way and got it working with both mutable and immutable entities. Check out #239 when you get a chance.

paularmstrong:
Merged #239 to handle this. Thanks for all the work to get this started, @andyhite

PR #239, opened 2017-02-16
Handle recursive relationships in denormalize

### LINKED ISSUE #216: Problem with denormalize method on recursive linked entities

jcarbo:
# Problem

Fixes #216 - denormalization with circular/recursive entities. Would replace #238.

# Solution

Turned `unvisit` into a factory function that returns a cache-aware version of the former unvisit. When `unvisiting` an entity, we first check if it's in the cache

Some other benefits:
- `Entity.denormalize` is simpler: given an entity object, denormalize it. Previously it would handle either an ID or an object, but this logic is moved up a level.
- Previously if you denormalized the same entity multiple times (e.g. array of books with the same author) the entities would not be referentially equal. This was due to how we were preventing mutating the entities in the store. With this PR, those entities will have the same object reference.
- No longer need to pass around `getDenormalizedEntities` through all of the denormalizers (it was only every used in the `Entity` schema.
- Works both for mutable and immutable. There's a limitation to the immutable functionality where when it reaches a circuluar dependency it will yield the original (normalized) entity instead of a circular reference. This is a limitation of immutable-js: https://github.com/facebook/immutable-js/issues/259#issuecomment-67899036

# TODO

- [x] Add & update tests

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -53,33 +53,69 @@ export const normalize = (input, schema) => {
   return { entities, result };
 };
 
-const unvisit = (input, schema, getDenormalizedEntity) => {
-  if (typeof schema === 'object' && (!schema.denormalize || typeof schema.denormalize !== 'function')) {
-    const method = Array.isArray(schema) ? ArrayUtils.denormalize : ObjectUtils.denormalize;
-    return method(schema, input, unvisit, getDenormalizedEntity);
+const unvisitEntity = (input, schema, unvisit, getEntity, cache) => {
+  const entity = getEntity(input, schema);
+  if (typeof entity !== 'object' || entity === null) {
+    return entity;
   }
 
-  if (input === undefined || input === null) {
-    return input;
+  const id = schema.getId(entity);
+
+  if (!cache[schema.key]) {
+    cache[schema.key] = {};
   }
 
-  return schema.denormalize(input, unvisit, getDenormalizedEntity);
+  if (!cache[schema.key][id]) {
+    // Ensure we don't mutate it non-immutable objects
+    const entityCopy = ImmutableUtils.isImmutable(entity) ? entity : { ...entity };
+
+    // Need to set this first so that if it is referenced further within the
+    // denormalization the reference will already exist.
+    cache[schema.key][id] = entityCopy;
+    cache[schema.key][id] = schema.denormalize(entityCopy, unvisit);
+  }
+
+  return cache[schema.key][id];
 };
 
-const getEntity = (entityOrId, schemaKey, entities, isImmutable) => {
-  if (typeof entityOrId === 'object') {
-    return entityOrId;
-  }
+const getUnvisit = (entities) => {
+  const cache = {};
+  const getEntity = getEntities(entities);
 
-  return isImmutable ?
-    entities.getIn([ schemaKey, entityOrId.toString() ]) :
-    entities[schemaKey][entityOrId];
+  const unvisit = (input, schema) => {
+    if (typeof schema === 'object' && (!schema.denormalize || typeof schema.denormalize !== 'function')) {
+      const method = Array.isArray(schema) ? ArrayUtils.denormalize : ObjectUtils.denormalize;
+      return method(schema, input, unvisit);
+    }
+
+    if (input === undefined || input === null) {
+      return input;
+    }
+
+    if (schema instanceof EntitySchema) {
+      return unvisitEntity(input, schema, unvisit, getEntity, cache);
+    }
+
+    return schema.denormalize(input, unvisit);
+  };
+
+  return unvisit;
 };
 
-const getEntities = (entities, isImmutable) => (schema, entityOrId) => {
-  const schemaKey = schema.key;
+const getEntities = (entities) => {
+  const isImmutable = ImmutableUtils.isImmutable(entities);
 
-  return getEntity(entityOrId, schemaKey, entities, isImmutable);
+  return (entityOrId, schema) => {
+    const schemaKey = schema.key;
+
+    if (typeof entityOrId === 'object') {
+      return entityOrId;
+    }
+
+    return isImmutable ?
+      entities.getIn([ schemaKey, entityOrId.toString() ]) :
+      entities[schemaKey][entityOrId];
+  };
 };
 
 export const denormalize = (input, schema, entities) => {
@@ -87,7 +123,5 @@ export const denormalize = (input, schema, entities) => {
     return input;
   }
 
-  const isImmutable = ImmutableUtils.isImmutable(entities);
-  const getDenormalizedEntity = getEntities(entities, isImmutable);
-  return unvisit(input, schema, getDenormalizedEntity);
+  return getUnvisit(entities)(input, schema);
 };
--- a/src/index.js
+++ b/src/index.js
@@ -82,7 +82,7 @@ const getUnvisit = (entities) => {
   const cache = {};
   const getEntity = getEntities(entities);
 
-  const unvisit = (input, schema) => {
+  return function unvisit(input, schema) {
     if (typeof schema === 'object' && (!schema.denormalize || typeof schema.denormalize !== 'function')) {
       const method = Array.isArray(schema) ? ArrayUtils.denormalize : ObjectUtils.denormalize;
       return method(schema, input, unvisit);
@@ -98,8 +98,6 @@ const getUnvisit = (entities) => {
 
     return schema.denormalize(input, unvisit);
   };
-
-  return unvisit;
 };
 
 const getEntities = (entities) => {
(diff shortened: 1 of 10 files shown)


src/schemas/Array.js
```diff
@@ -21,10 +21,10 @@ export const normalize = (schema, input, parent, key, visit, addEntity) => {
   return values.map((value, index) => visit(value, parent, key, schema, addEntity));
 };
 
-export const denormalize = (schema, input, unvisit, getDenormalizedEntity) => {
+export const denormalize = (schema, input, unvisit) => {
   schema = validateSchema(schema);
-  return Array.isArray(input) ?
-    input.map((entityOrId) => unvisit(entityOrId, schema, getDenormalizedEntity)) :
+  return (input && input.map) ?
```
jcarbo: `Array.isArray` returns false for immutable lists. This change makes it work for both mutable `Array`s and immutable `List`s and doesn't break any existing tests.

src/schemas/Array.js
```diff
@@ -36,9 +36,9 @@ export default class ArraySchema extends PolymorphicSchema {
       .filter((value) => value !== undefined && value !== null);
   }
 
-  denormalize(input, unvisit, getDenormalizedEntity) {
-    return Array.isArray(input) ?
-      input.map((value) => this.denormalizeValue(value, unvisit, getDenormalizedEntity)) :
+  denormalize(input, unvisit) {
+    return (input && input.map) ?
```
jcarbo: Same note above about immutable lists.

src/schemas/Entity.js
```diff
@@ -61,23 +61,17 @@ export default class EntitySchema {
...
-      return entity;
-    }
-
+  denormalize(entity, unvisit) {
     if (ImmutableUtils.isImmutable(entity)) {
-      return ImmutableUtils.denormalizeImmutable(this.schema, entity, unvisit, getDenormalizedEntity);
+      return ImmutableUtils.denormalizeImmutable(this.schema, entity, unvisit);
     }
 
-    const processedEntity = { ...entity };
```
jcarbo: This was where we were preventing mutations of the entities in the store which was creating a new reference each time an entity was denormalized. This is now just done once when an entity is first "unvisited", so any subsequent hits will have the same exact reference.

src/index.js
```diff
@@ -53,41 +53,75 @@ export const normalize = (input, schema) => {
...
+  return cache[schema.key][id];
 };
 
-const getEntities = (entities, isImmutable) => (schema, entityOrId) => {
-  const schemaKey = schema.key;
+const getUnvisit = (entities) => {
+  const cache = {};
+  const getEntity = getEntities(entities);
+
+  const unvisit = (input, schema) => {
```
paularmstrong: nit: can you return this function directly instead of assigning to a `const`. Makes it clearer that no other work will happen in `getUnvisit`

src/schemas/__tests__/Entity.test.js
```diff
@@ -201,4 +201,40 @@ describe(`${schema.Entity.name} denormalization`, () => {
...
+          role: 'manager',
+          reports: [ '123' ]
+        }
+      }
+    };
+    expect(denormalize('123', report, entities)).toMatchSnapshot();
+    expect(denormalize('123', report, fromJS(entities))).toMatchSnapshot();
+
+    expect(denormalize('456', user, entities)).toMatchSnapshot();
+    expect(denormalize('456', user, fromJS(entities))).toMatchSnapshot();
```
paularmstrong: Can you add (non-immutable) assertions that show the recursive entities are the same as well? It'll help in case the snapshost is incorrectly updated.

```
expect(denormalized).toBe(denormalized.draftedBy.reports[0]);
```

src/schemas/__tests__/__snapshots__/Entity.test.js.snap
```diff
@@ -92,6 +92,100 @@ Object {
...
+        "title": "Weekly report",
+      },
+    ],
+    "role": "manager",
+  },
+  "id": "123",
+  "publishedBy": Object {
+    "id": "456",
+    "reports": Array [
+      Object {
```
paularmstrong: Why do Immutables go an extra level deep instead of stopping at the first duplication? I'd expect this to just be `"123"` instead of a repeat of the original object.

paularmstrong:
@jcarbo this looks really great!
@andyhite would you mind taking a look as well?

andyhite:
@jcarbo this looks awesome! Turning unvisit into a factory is a great idea. Thanks @jcarbo and @paularmstrong!

jcarbo:
Thanks for the review! I'm traveling today but I'll address those comments later tonight when I'm back at a computer.

src/index.js
jcarbo: Given that we need to recursively call `unvisit` within itself it needs to be named so we can't return an anonymous function. However, we can just return a named function which I think addresses this.

src/schemas/__tests__/__snapshots__/Entity.test.js.snap
jcarbo: When denormalizing entities, we're essentially doing a depth first search until we run out of nested schemas or we hit an entity we've already seen. To avoid infinite loop, we store the reference to the original (yet-to-be-denormalized) entity in the cache before recursing further.

For mutable objects, this results in us using the same exact reference which we're just mutating in `Entity.denormalize` and so creates a circular reference.
```js
report.draftedBy.reports[0] = originalReport
report === originalReport
```

For immutable objects, we point to the original reference, but during denormalization the entity's reference is changing.
```js
report.draftedBy.reports[0] = originalReport
report !== originalReport
```

src/schemas/__tests__/Entity.test.js
jcarbo: Added an additional spec for this.

jcarbo:
@paularmstrong - added an additional test and returned the `unvisit` function directly as you suggested. Should be g2g 👍

src/schemas/__tests__/Entity.test.js
```diff
@@ -201,4 +201,79 @@ describe(`${schema.Entity.name} denormalization`, () => {
...
+      }
+    };
+
+    const denormalizedReport = denormalize('123', report, entities);
+
+    expect(denormalizedReport).toBe(denormalizedReport.draftedBy.reports[0]);
+    expect(denormalizedReport.publishedBy).toBe(denormalizedReport.draftedBy);
+
+    // NOTE: Given how immutable data works, referential equality can't be
+    // maintained with nested denormalization.
```
paularmstrong: Thanks for the comment!

src/schemas/__tests__/__snapshots__/Entity.test.js.snap
paularmstrong: Good to know. Thanks!

andyhite:
I've been keeping up with @jcarbo's work on this as he was pushing stuff and it all looks great. Excited to see this in the next release!

paularmstrong:
Thanks for all the hard work and collaboration @andyhite, @jcarbo!

fxck:
@paularmstrong any idea when you could merge and release this?

fx:
+1 for releasing, works fine.

paularmstrong:
Now available in v3.2.2 via npm

swieder227:
Question on the API documentation. 

Currently reads:
> If your schema and data have recursive references, only the first instance of an entity will be given. **Subsequent references will be returned as the id provided.**

Is that last line still accurate? It seems like instead of returning the `id`, it returns a reference to the first entity instance.

sarink:
Just a thought... why not use ES6 getters for this to lazily evaluate the reference? That way the API remains consistent, and circular references would be avoided

Issue #243, opened 2017-02-23
Normalizing recursive payload doesn't work


GertjanReynaert:
# Problem

Normalizing recursive payload gives an error instead of normalizing the payload.

(I've seen some work on recursive issues in denormalize, but I wasn't sure if you were aware of this issue when normalizing payload, that's why I created this issue)

**Input**

I created a failing spec to best illustrate the issue:
https://github.com/GertjanReynaert/normalizr/blob/gr-failing-recursive-normalize/src/__tests__/index.test.js#L53

**Output**

Instead of getting a normalized output, I get the following error:
```
 FAIL  src/__tests__/index.test.js
  ● normalize › normalizes recursive payloads

    TypeError: Cannot read property 'normalize' of undefined

      at visit (src/index.js:18:16)
      at src/schemas/Array.js:21:79
      at Array.map (native)
      at normalize (src/schemas/Array.js:21:17)
      at visit (src/index.js:15:12)
      at src/schemas/Entity.js:56:34
      at Array.forEach (native)
      at EntitySchema.normalize (src/schemas/Entity.js:53:32)
      at visit (src/index.js:18:17)
      at src/schemas/Array.js:21:79
      at Array.map (native)
      at normalize (src/schemas/Array.js:21:17)
      at visit (src/index.js:15:12)
      at normalize (src/index.js:52:16)
      at Object.<anonymous> (src/__tests__/index.test.js:81:55)
```

paularmstrong:
#216, #239

GertjanReynaert:
@paularmstrong I was aware of these two issues, but these are related to `DEnormalizing`. I'm not sure if the issue I have with `normalizing` is related to those denormalizing issues.

paularmstrong:
Uh... normalizing recursive? My initial reaction is... don't do that. Normalizr was written for JSON data, which cannot be recursive.

GertjanReynaert:
@paularmstrong I agree with your initial reaction, but I can't change the payload on our api so that's not really an option.

The recursive normalization as shown in my [failing spec](https://github.com/paularmstrong/normalizr/pull/244) worked pre 3.x, but in 3.x it no longer does.

This isn't that big of an issue to me, since I can work around it locally, but I thought you might want to know about this regression.

PR #244, opened 2017-02-23
Add failing spec for normalizing recursive payload.

GertjanReynaert:
Pr with failing spec for issue https://github.com/paularmstrong/normalizr/issues/243

### DIFF
--- a/src/__tests__/index.test.js
+++ b/src/__tests__/index.test.js
@@ -50,6 +50,37 @@ describe('normalize', () => {
     expect(normalize(input, article)).toMatchSnapshot();
   });
 
+  it('normalizes recursive payloads', () => {
+    const user = new schema.Entity('users', {
+      comments: [comment],
+    });
+
+    const comment = new schema.Entity('comments', {
+      user: user,
+    });
+
+    const userList = [user];
+
+    const input = [
+      {
+        id: '123',
+        name: 'paul',
+        comments: [
+          {
+            id: '456',
+            text: 'Test comment',
+            user: {
+              id: '123',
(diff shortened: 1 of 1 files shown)


src/__tests__/index.test.js
```diff
@@ -50,6 +50,37 @@ describe('normalize', () => {
     expect(normalize(input, article)).toMatchSnapshot();
   });
 
+  it('normalizes recursive payloads', () => {
+    const user = new schema.Entity('users', {
+      comments: [comment],
```
paularmstrong: `comment` is undefined here.

You just need to switch this to...

```
    const user = new schema.Entity('users');
    const comment = new schema.Entity('comments', {
      user: user,
    });
    user.define({
      comments: [ comment ]
    });
```

paularmstrong:
In your issue, I thought you were talking about actual recursive data (where data[0] === data[0].comments[0].user). You just need to ensure what you're defining to your schema is actually a valid `Entity` instead of `undefined`

src/__tests__/index.test.js
GertjanReynaert: @paularmstrong I agree that in this case this is indeed the solution. However in my project I have `schemas/user.js` and `schemas/comment.js`.

How can I then solve the following:
```
// user.js
import { schema } from 'normalizr';
import { comment } from './comment.js';

export const user = new schema.Entity('users', {
  comments: [comment],
});
```
and 
```
// comment.js
import { schema } from 'normalizr';
import { user } from './user.js';

export const comment = new schema.Entity('comments', {
  user: user,
});
```

PR #252, opened 2017-03-21
Add tests for polymorphic collections of entities

azangru:
# Problem

This pull requests supplies two failing test cases when data normalized using array schemas is denormalized incorrectly when using the `schema.Array` constructor, but denormalizes correctly when using simple javascript arrays.

## Case 1
Sample data is an array of entities. It was normalized in an ordinary way (`normalize(data, new schema.Array(schema))`. When denormalizing, this approach works: `denormalize(result, [ schema ], entities)`, while this doesn’t: `denormalize(result, new schema.Array(schema), entities)`

Difference between expected and actual sample data:

image

## Case 2
Sample data is a record with a field that contains an array schema. If the schema was constructed using simple javascript array:
```
schema.define({
    field: [ anotherSchema ]
});
```
denormalization works, whereas if the schema was created using the `schema.Array` consctructor:
```
schema.define({
    field: new schema.Array(anotherSchema)
});
```
denormalization fails. Difference between expected and actual sample data:

image

### DIFF
--- a/src/__tests__/index.test.js
+++ b/src/__tests__/index.test.js
@@ -187,6 +187,75 @@ describe('denormalize', () => {
     expect(denormalize('123', article, entities)).toMatchSnapshot();
   });
 
+  it('denormalizes arrays of entities normalized using Union schemas)', () => {
+    const admin = {
+      type: 'admin',
+      id: 1,
+      name: 'Mr. Admin'
+    };
+    const visitor = {
+      type: 'visitor',
+      id: 2,
+      name: 'Ms Visitor'
+    };
+    const adminSchema = new schema.Entity('admin');
+    const visitorSchema = new schema.Entity('visitor');
+    const userSchema = new schema.Union({
+      admin: adminSchema,
+      visitor: visitorSchema
+    }, 'type');
+    const data = [ admin, visitor ];
+    const { result, entities } = normalize(data, new schema.Array(userSchema));
+
+    // example when schema for denormalization is created using a javascript array
+    const denormalizedData1 = denormalize(result, [ userSchema ], entities);
+    // example when schema for denormalization is created using schema.Array constructor
+    const denormalizedData2 = denormalize(result, new schema.Array(userSchema), entities);
+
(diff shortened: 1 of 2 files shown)


src/__tests__/index.test.js
```diff
@@ -187,6 +187,75 @@ describe('denormalize', () => {
...
+      name: 'Ms Visitor'
+    };
+    const adminSchema = new schema.Entity('admin');
+    const visitorSchema = new schema.Entity('visitor');
+    const userSchema = new schema.Union({
+      admin: adminSchema,
+      visitor: visitorSchema
+    }, 'type');
+    const data = [ admin, visitor ];
+    const { result, entities } = normalize(data, new schema.Array(userSchema));
```
paularmstrong: Instead of using `userSchema = new schema.Union`, you should only be using `schema.Array`:

```
const userListSchema = new schema.Union({
  admin: adminSchema,
  visitor: visitorSchema
}, 'type');

const { result, entities } = normalize(data, userListSchema);

const denormalized = denormalize(result, userListSchema, entities);
```

The reason that `[ unionSchema ]` works is because plain array syntax does not allow polymorphic mappings, while `schema.Array` does. 

Internally, if you use `schema.Array` with a `schema.Union` as its internal schema, you end up nesting polymorphic mappings, so resolving the schema for a single object in the input array (`result`, in the test code) resolves to a mapping of multiple potential types. `schema.Array` has this mapping potential already built-in to itself.

azangru:
@paularmstrong Oh, I see, I misread the docs about `schema.Array`  — I thought a polymorphic collection was a collection of polymorphic schemas (each of which is defined by `schema.Union`), while in fact `schema.Array` is entirely capable of creating a polymorphic collection by itself.

I went ahead and fixed the tests to prove to myself that your suggestion works. Don’t know if you would like to add these tests to the test suite. If not, feel free to close the PR.

Thank you for the clarification.

paularmstrong:
Great! If you can think of a way to make the docs more clear, I'd be happy to look at a PR for that!

PR #254, opened 2017-04-03
fix(definitions): Fixed Entity#key definition for typescript

Mordred:
# Wrong definition for Entity#key

`Entity.key` property is defined using getter (https://github.com/paularmstrong/normalizr/blob/master/src/schemas/Entity.js#L28), but in the definition there is a `key(): string` method (https://github.com/paularmstrong/normalizr/blob/master/index.d.ts#L20) which is wrong.

# Solution

Fixed index.d.ts based on source code: https://github.com/paularmstrong/normalizr/blob/master/src/schemas/Entity.js#L28

# TODO

- [x] Add & update tests

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -17,7 +17,7 @@ declare namespace schema {
   export class Entity {
     constructor(key: string, definition?: Schema, options?: EntityOptions)
     define(definition: Schema): void
-    key(): string
+    key: string
   }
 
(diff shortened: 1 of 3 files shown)


paularmstrong:
Can you please add a test to that fails without this change?

Mordred:
@paularmstrong Added

Issue #255, opened 2017-04-19
denormalize using `schema.Union` with Immutable data does not denormalize


n-sviridenko:
# Problem

It doesn't denormalize the entity.

**Input**

Here's how I'm using normalizr:

short:
https://runkit.com/58f68083e5dc270012c70492/58f7a233718530001238d7b9

long:
```js
var { schema, denormalize } = require("normalizr");
var { fromJS } = require("immutable");

const a = new schema.Entity('as');
const b = new schema.Entity('bs');
const ab = new schema.Union({ a, b }, 'type');

const loader = {
    loading: false,
    error: null,
    data: {
        id: 1,
        schema: 'a',
    },
};

const entities = {
    as: {
        '1': {
            id: 1,
            type: 'a',
        },
    },
};

console.log(denormalize(loader, { data: ab }, entities));
console.log(denormalize(fromJS(loader), { data: ab }, fromJS(entities)).toJS());
```

**Output**

```js
{
	"loading": false,
	"error": null,
	"data": {
		"id": 1,
		"type": "a"
	}
}
{
	"loading": false,
	"error": null,
	"data": {
		"id": 1,
		"schema": "a"
	}
}
```

n-sviridenko:
Hello! Is there any news?

paularmstrong:
Fixed by #257

PR #257, opened 2017-05-01
Fixed denormalization using `schema.Union` with immutable data

n-sviridenko:
# Problem
https://github.com/paularmstrong/normalizr/issues/255

Explain the problem that this pull request aims to resolve.

# Solution
Implemented

Explain your approach. Sometimes it helps to justify your approach against some others that you didn't choose to explain why yours is better.

# TODO

- [ ] Add & update tests
- [ ] ~Update relevant documentation~ (no need)

### DIFF
--- a/src/schemas/Polymorphic.js
+++ b/src/schemas/Polymorphic.js
@@ -43,11 +43,12 @@ export default class PolymorphicSchema {
   }
 
   denormalizeValue(value, unvisit) {
-    const mutableValue = isImmutable(value) ? value.toJS() : value;
-    if (!this.isSingleSchema && !mutableValue.schema) {
+    const schemaKey = isImmutable(value) ? value.get('schema') : value.schema;
+    if (!this.isSingleSchema && !schemaKey) {
       return value;
     }
-    const schema = this.isSingleSchema ? this.schema : this.schema[mutableValue.schema];
-    return unvisit(mutableValue.id || value, schema);
+    const id = isImmutable(value) ? value.get('id') : value.id;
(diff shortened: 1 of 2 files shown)


n-sviridenko:
Ping @paularmstrong

src/schemas/Polymorphic.js
```diff
@@ -41,10 +43,11 @@ export default class PolymorphicSchema {
   }
 
   denormalizeValue(value, unvisit) {
-    if (!this.isSingleSchema && !value.schema) {
+    const mutableValue = isImmutable(value) ? value.toJS() : value;
```
paularmstrong: Do we have to convert `toJS()`? My (limited) understanding is that it's expensive. Can't we use `mutableValue.get('schema')`?

jeffcarbs: Yea it can be expensive. From [the docs](https://facebook.github.io/immutable-js/docs/#/Map/toJS) for `.toJS()`:
> Deeply converts this Keyed collection to equivalent native JavaScript Object.

Since we only care about accessing the `schema` and `id` key here I think this would be better written as:
```js
const schemaKey = isImmutable(value) ? value.get('schema') : value.schema;
if (!this.isSingleSchema && !schemaKey) {
  return value;
}
const schema = this.isSingleSchema ? this.schema : this.schema[schemaKey];
const id = isImmutable(value) ? value.get('id') : value.id;
return unvisit(id || value, schema);
```

n-sviridenko:
Ping @paularmstrong

Issue #260, opened 2017-05-10
idAttribute as function behaves differently than as string [doc?]


Stenerson:
# Problem

It's not obvious when reading the [API documentation](https://github.com/paularmstrong/normalizr/blob/master/docs/api.md#schema) that normalizr will behave differently when using `idAttribute` as a function vs. a string. Based on [the tests](https://github.com/paularmstrong/normalizr/blob/master/src/schemas/__tests__/Entity.test.js#L96), I assume this behavior is intentional but it caused some heartache for me when attempting some non-standard normalizations.

### As String
**Input Data**
```js
const myData = {
  patrons: [
    {
      patron_id: 1,
      name: "Esther",
      ...other fields that actually require normalization
    },
    {
      patron_id: 2,
      name: "Tom",
      ...other fields that actually require normalization
    },
  ]
};
```

In this case we're don't have an `id` field so we need to tell normalizr to use `patron_id`, no problem:

```js
const patronsSchema = new schema.Entity('patrons', undefined, {
  // idAttribute *strings* must return the id's **key**
  idAttribute: 'patron_id',
});

normalize(myData, [patronsSchema]);

// Result
{
    result: [1, 2],
    entities: {
      patrons: {
        1: { ... },
        2: { ... },
      }
    }
}
```

### As Function
**Input Data**
```js
const myData = {
  patrons: [
    {
      id: 1,
      guest_id: null,
      name: "Esther",
      ...other fields that actually require normalization
    },
    {
      id: 1,
      guest_id: 22,
      name: "Tom",
      ...other fields that actually require normalization
    },
  ]
};
```

In this example we have a composite key `id + guest_id`. Because of the way `idAttribute` works as a function we can create a composite key value to uniquify them.

```js
const patronsSchema = new schema.Entity('patrons', undefined, {
  // idAttribute *functions* must return the id's **value** (not key)
  idAttribute: value => value.guest_id ? `${value.id}-${value.guest_id}` : value.id,
});

normalize(myData, [patronsSchema]);

// Result
{
    result: [1, 1-22],
    entities: {
      patrons: {
        1: { ... },
        1-22: { ... },
      }
    }
}
```

# Suggestions

Currently the doc states:

- `options`:
    - `idAttribute`: The attribute where unique IDs for each of this entity type can be found. Accepts either a string key or a function. Defaults to id. 
    As a function, accepts the following arguments, in order:
        - `value`: The input value of the entity.
        - `parent`: The parent object of the input array.
[...]

Issue #258, opened 2017-05-08
Composite/Calculated ID field requires input mutation


Stenerson:
# Problem

Adding a composite ID in `processStrategy` is not being transferred to the normalized entity correctly unless I mutate the `value` within `processStrategy`.

**Input**

Here's how I'm using normalizr:

```js
const myData = {
  patrons: [
    {
      id: 1,
      name: "Esther",
      guest_id: null,
      ...other fields that actually require normalization
    },
    {
      id: 1,
      name: "Tom",
      guest_id: 22,
      ...other fields that actually require normalization
    },
  ]
};

const patronProcessStrategy = (value, parent, key) => {
  // Determine composite key
  const compositeId = value.guest_id ? `${value.id}-${value.guest_id}` : value.id;
  // Return a non-mutated version of the value with the key added
  return { ...value, compositeId };
};
const patronsSchema = new schema.Entity('patrons', undefined, {
  idAttribute: 'compositeId',
  processStrategy: patronProcessStrategy,
});

normalize(myData, [patronsSchema]);
```

**Output**

Here's what I expect to see when I run the above:

```js
{
    result: [1, 1-22],
    entities: {
      patrons: {
        1: { ... },
        1-22: { ... },
      }
    }
}
```

Here's what I *actually* see when I run the above:

```js
{
    result: [undefined, undefined],
    entities: 
      patrons: { undefined: { ... } }
}
```

**Possible Solution**
If I mutate the value in `processStrategy` **I get the output I expect**. This approach was [*kind of* validated](https://github.com/paularmstrong/normalizr/issues/256#issuecomment-298128047) in #256 but generally, mutation is bad and it seems especially so here.

```js
const patronProcessStrategy = (value, parent, key) => {
  // Determine composite key
  const compositeId = value.guest_id ? `${value.id}-${value.guest_id}` : value.id;
  // Mutate!
  value.compositeId = compositeId;
  // Return copy of the (already mutated object)
  return { ...value };
};
```

Based on the [suggestions](https://github.com/paularmstrong/normalizr/blob/master/docs/api.md#entitykey-definition---options--), I'd like to avoid mutating the object. I can see from the [source](https://github.com/paularmstrong/normalizr/blob/2d60121de8810e6694a353c815f4f93a9424ef5d/src/index.js#L23) that the original input object is used to actually get the entity's ID for the normalized output. I'm not sure if this is intentional or a bug but I'd think that we should be able to use the `processedEntity` to get the key here.

[...]

PR #261, opened 2017-05-13
Use processedEntity to get idAttribute

Stenerson:
# Problem

Changes made to the object in `processStrategy` are not available to the `idAttribute` unless the value is mutated in `processStrategy`.

See discussion in #258 

# Solution

Change source to use `processedEntity` to get the ID, rather than the non-normalized value. This allows us to get the ID using a string rather than a function without mutating the input value.

## Alternative Solution

Don't mutate the input value and pass a function to `idAttirbute` that returns the computed ID **value**.

# TODO

- [x] Add & update tests
- [ ] Ensure CI is passing (lint, tests, flow) **Requires discussion**
- [x] Update relevant documentation **none?**

# Issues
There is [one test](https://github.com/paularmstrong/normalizr/blob/master/src/__tests__/index.test.js#L108) that's failing due to this change. Since **this is a breaking change**, I thought we could discuss the possible solutions here; before I changed the intent of the library/tests.

```js
  it('uses the non-normalized input when getting the ID for an entity', () => {
    const userEntity = new schema.Entity('users');
    const idAttributeFn = jest.fn((nonNormalized, parent, key) => nonNormalized.user.id);
    const recommendation = new schema.Entity('recommendations', { user: userEntity }, {
      idAttribute: idAttributeFn
    });
    expect(normalize({ user: { id: '456' } }, recommendation)).toMatchSnapshot();
    expect(idAttributeFn.mock.calls).toMatchSnapshot();
    expect(recommendation.idAttribute).toBe(idAttributeFn);
  });
```

This change goes pretty much against the intent of this test since it's clearly now passing the partially normalized output to the `getId` function:

```js
  // Entity.js:51
  normalize(input, parent, key, visit, addEntity) {
    const processedEntity = this._processStrategy(input, parent, key);
    Object.keys(this.schema).forEach((key) => {
      if (processedEntity.hasOwnProperty(key) && typeof processedEntity[key] === 'object') {
        const schema = this.schema[key];
        // HERE, we're normalizing child schemas before passing to getId
        processedEntity[key] = visit(processedEntity[key], processedEntity, key, schema, addEntity);
      }
    });

    addEntity(this, processedEntity, input, parent, key);
    // Here's the only change I made, passing processedEntity, vs input
    return this.getId(processedEntity, parent, key);
  }
```

I see a couple options to move forward. Please let me know which one you'd prefer @paularmstrong
- If you want to move forward with this, we can discuss options en route to a breaking change (also [see additional discussion here](https://github.com/paularmstrong/normalizr/issues/258#issuecomment-300592753))
- Close this, in favor of simply updating the doc --> #260

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -20,7 +20,7 @@ const visit = (value, parent, key, schema, addEntity) => {
 
 const addEntities = (entities) => (schema, processedEntity, value, parent, key) => {
   const schemaKey = schema.key;
-  const id = schema.getId(value, parent, key);
+  const id = schema.getId(processedEntity, parent, key);
   if (!(schemaKey in entities)) {
     entities[schemaKey] = {};
   }
--- a/src/schemas/Entity.js
+++ b/src/schemas/Entity.js
@@ -58,7 +58,7 @@ export default class EntitySchema {
     });
 
     addEntity(this, processedEntity, input, parent, key);
-    return this.getId(input, parent, key);
+    return this.getId(processedEntity, parent, key);
   }
 
   denormalize(entity, unvisit) {
--- a/src/schemas/__tests__/Entity.test.js
+++ b/src/schemas/__tests__/Entity.test.js
@@ -93,9 +93,21 @@ describe(`${schema.Entity.name} normalization`, () => {
       // If not run before, this schema would require a parent object with key "message"
       const myEntity = new schema.Entity('entries', {
         data: { attachment: attachmentEntity }
-      }, { idAttribute: (input) => Object.values(input)[0].id, processStrategy });
+      }, { idAttribute: (input) => input.id, processStrategy });
 
       expect(normalize({ message: { id: '123', data: { attachment: { id: '456' } } } }, myEntity)).toMatchSnapshot();
+    });
+
+    it('is run before and passed to the schema normalization', () => {
+      const processStrategy = (input) => ({ ...input, compId: `${input.id}-${input.sid}` });
+      // If not run before, compId wouldn't exist
+      const myEntity = new schema.Entity('patrons', undefined,
+        { idAttribute: 'compId', processStrategy }
+      );
+
+      expect(
+        normalize([ { id: '123', sid: '456' }, { id: '123', sid: '789' } ], [ myEntity ])
+      ).toMatchSnapshot();
     });
   });
 });
--- a/src/schemas/__tests__/__snapshots__/Entity.test.js.snap
+++ b/src/schemas/__tests__/__snapshots__/Entity.test.js.snap
@@ -370,3 +370,26 @@ Object {
   "result": "123",
 }
 `;
+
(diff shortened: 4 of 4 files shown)


paularmstrong:
> There is one test that's failing due to this change. Since this is a breaking change, I thought we could discuss the possible solutions here; before I changed the intent of the library/tests.

Ah, yeah... that's actually a very important feature. If we're going to make a big breaking change like this, we should probably refactor the `idAttribute` function to take both the processedEntity and the input data

Stenerson:
That's a good idea. Adding `processedEntity` as a 4th parameter is a little ugly (because it's out of order from other uses) but it does make it a non-breaking change. 👍 

With that being said, I'm starting to lean towards just clarifying the doc and leaving it at that. I've updated my app to use a shared function to create a 'compositeId' in `processedEntity` and `idAttribute` and it's working great. 

Once I understood the intent of `idAttribute` I was able to do all I needed to with the current functionality. 

Can you think of other uses for this feature?

paularmstrong:
> Can you think of other uses for this feature?

Not really. It's nice to have, but as you said, it really just bloats things a little bit since all of the data is actually readily available to you by passing the `input`, `key`, and `parent` values through.

I'll close this for now. Please reopen if you find a compelling case that contradicts our current arguments and think we need this.

newtriks:
This would be useful for an instance I have come up against recently where an API is not generating a unique `id` and expects the front end to update the `id` with a unique value. 

I am now sat with a predicament that I need to (a) set the `idAttribute` to a unique `id` and (b) apply the same value to the object `id`. 

Both `processStrategy` and `idAttribute` have no way of passing the uniquely generated `id` to each other. 

I could follow the `idAttribute` step and then manipulate the data post process outside of the schema, however, it would be nice to do it all in one hit so to speak.

Stenerson:
I could see this being an issue if your client-generated ID is not derived in any way from the data. Is it possible to derive the ID from the data? **If you can** use the data to generate your ID, [this solution](https://github.com/paularmstrong/normalizr/issues/258#issuecomment-300268601) may work for you.

newtriks:
Unfortunately that wouldn't work, this is unusual I know, but it's entirely feasible for there to be multiple objects that are identical... :(

paularmstrong:
Unfortunately, generating a UUID without using the api-provided data is going to be a really bad practice, because you won't have a way to resolve getting the same entity more than once in either the same cycle or another cycle, so you'll always have a new entity, even though the data is the same. 

If the objects are guaranteed to be identical, you could use a hashing function, though it'll be a bit expensive:

```
const hashCode = (str) {
  let hash = 0;
  for (let i = 0; i < str.length; i++) {
    const chr   = str.charCodeAt(i);
    hash  = ((hash << 5) - hash) + chr;
    hash |= 0; // Convert to 32bit integer
  }
  return hash;
};
const entityHash = hashCode(JSON.stringify(myEntity));
```

newtriks:
Yeah, I agree. Request for API change has been made 😛. Thanks!

PR #262, opened 2017-05-17
DOC: Clarify Entity idAttribute behavior

### LINKED ISSUE #260: idAttribute as function behaves differently than as string [doc?]

Stenerson:
# Problem

Doc is unclear what the return value of `idAttribute` function should be.

Closes #260 

# Solution

Update the doc to clarify, also include an example.

# TODO

- [X] Add & update tests **N/A**
- [X] Ensure CI is passing (lint, tests, flow) **N/A**
- [X] Update relevant documentation

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -216,37 +216,37 @@ const normalizedData = normalize(data, tweet);
 }
 ```
 
+#### `idAttribute` Usage
+
 When passing the `idAttribute` a function, it should return the IDs value. 
 
 For Example:
 
 ```js
-const myData = {
-  patrons: [
-    { id: 1, guest_id: null, name: "Esther" },
-    { id: 1, guest_id: 22, name: "Tom" },
-  ]
-};
+const data = [
+    { id: '1', guest_id: null, name: 'Esther' },
+    { id: '1', guest_id: '22', name: 'Tom' },
+];
 
 const patronsSchema = new schema.Entity('patrons', undefined, {
   // idAttribute *functions* must return the ids **value** (not key)
   idAttribute: value => value.guest_id ? `${value.id}-${value.guest_id}` : value.id,
 });
 
-normalize(myData, [patronsSchema]);
+normalize(data, [patronsSchema]);
 ```
 
 #### Output
 
 ```js
 {
-    entities: {
-      patrons: {
-        1: { ... },
-        1-22: { ... },
-      }
-    },
-    result: [1, 1-22]
+  entities: {
+    patrons: {
+      '1': { id: '1', guest_id: null, name: 'Esther' },
+      '1-22': { id: '1', guest_id: '22', name: 'Tom' },
+    }
+  },
+  result: ['1', '1-22']
 }
 ```
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -164,7 +164,7 @@ const normalizedData = normalize(data, myArray);
 You *do not* need to define any keys in your entity other than those that hold nested entities. All other values will be copied to the normalized entity's output.
 * `options`:
     - `idAttribute`: The attribute where unique IDs for each of this entity type can be found.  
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -216,6 +216,41 @@ const normalizedData = normalize(data, tweet);
...
+```
+
+#### Output
+
+```js
+{
+    entities: {
+      patrons: {
+        1: { ... },
+        1-22: { ... },
```
paularmstrong: nit: make sure these are quoted as strings, otherwise that's invalid JS, ex: `'1-22': { ... }`

docs/api.md
```diff
@@ -216,6 +216,41 @@ const normalizedData = normalize(data, tweet);
...
+});
+
+normalize(myData, [patronsSchema]);
+```
+
+#### Output
+
+```js
+{
+    entities: {
```
paularmstrong: nit: 2-space indents, please

docs/api.md
```diff
@@ -216,6 +216,41 @@ const normalizedData = normalize(data, tweet);
 }
 ```
 
+When passing the `idAttribute` a function, it should return the IDs value.
```
paularmstrong: Can you add an h4 above this?

```md
#### `idAttribute` Usage
```

Issue #206, opened 2017-01-04
Post-process data after normalization


dferber90:
# Problem
In normalizr v2 we were able to use the `assignEntity` function to modify data after it has been normalized. In v3 this functionality went away and we can only pre-process it. Would you accept a PR that adds functionality which lets us modify the entities after they have been normalized (but before they are added into the entities of the result)?

## Use case
Our use-case is very specific. I'll try to dumb it down: We need to remove some parts of each entity before the normalization and add it back in after the normalization. We can't keep it in during the normalization because then it is not possible to create a proper schema.

<details>
<summary>The nasty details, feel free to skip them because they can cause headaches (left this here mostly for my team members)</summary>

A product we get from the API (two possibilities here):
```js
// with an expanded category reference
{
  id: 'product-1',
  name: 'Sponge',
  category: { typeId: 'category', id: 'category-1', obj: { id: 'category-1', /* other category data */ } }
}

// with a non-expanded category reference
{
  id: 'product-1',
  name: 'Sponge',
  category: { typeId: 'category', id: 'category-1' }
}
```

In the first case we want the resulting entities to be:

```js
{
  results: 'product-1',
  entities: {
    products: {
      'product-1': { id: 'product-1', name: 'Sponge', category: { typeId: 'category', id: 'category-1' } },
    },
    categories: {
      'category-1': { id: 'category-1', /* other category data */  }
    }
  }
}
```

In the second case it would be 

```js
{
  results: 'product-1',
  entities: {
    products: {
      'product-1': { id: 'product-1', name: 'Sponge', category: { typeId: 'category', id: 'category-1' } },
    },
    categories: {
      'category-1': { id: 'category-1' }
    }
  }
}
```

Notice how we'd need to define the category schema on the `obj`? But what would happen if there is no `obj` like in the second case? Another problem here is that the category is contained within `obj`, but we can't define the schema there because then the reference after the normalization would end up in the wrong spot (`{ typeId: 'category', id: 'category-1', obj: 'category-1' }`, but it should be `{ typeId: 'category', id: 'category-1' }`. 

The solution we'd like to use is to preprocess the data to be able to transform it:

```js
// before preprocessing
{
  id: 'product-1',
  name: 'Sponge',
  category: { typeId: 'category', id: 'category-1' }
}

// after preprocessing (what the normalization will see)
{
  id: 'product-1',
  name: 'Sponge',
  category: { id: 'category-1' }
}
```

Now we can just define category to be a reference.

[...]

Issue #199, opened 2017-01-02
Union schemaAttribute function not called when process strategy returns immutablejs record


kolpav:
Does not call `schemaAttribute` function
```javascript
const removalLog = new schema.Entity(
  'removalLogs',
  {
    item: new schema.Union({
      post,
      comment
    }, (value, parent) => parent.itemType) // not called
  },
  {
    idAttribute: 'id',
    processStrategy: entity => new RemovalLogRecord(entity)
  }
)
```
Calls `schemaAttribute` function
```javascript
const removalLog = new schema.Entity(
  'removalLogs',
  {
    item: new schema.Union({
      post,
      comment
    }, (value, parent) => parent.itemType) // called
  },
  {
    idAttribute: 'id',
    processStrategy: entity => entity
  }
)
```

I would like to hold all my entites as immutable js records but it is not possible when using `schemaAttribute` function as it is never called for some reason.

paularmstrong:
I suspect this is a problem with your `processStrategy` for you `removalLogs` Entity. As described in the docs, this is done before visiting the properties as defined by the schema. If `RemovalLogRecord` process strategy removes that value from the Entity, then it won't be visited or normalized.

> Strategy to use when pre-processing the entity. Use this method to add extra data, defaults, and/or completely change the entity before normalization is complete.

kolpav:
What a quick response.
If you are faimiliar with `immutablejs` you propably know what I am trying to achieve basically conversion from js object to Records at the time of normalization. Previously I had ugly middleware for that. So I guess I am using `processStrategy` for something that it was not intended to, because if I convert it to `Record` there normalization logic then works with `Record` and not js object right? Do you think there could be similar callback as `processStrategy` but called after normalization? That would be a perfect place to convert to `Record` and I assume it would be very common use case.
Bellow is something that I also tried.
```javascript
const removalLog = new schema.Entity(
  'removalLogs',
  {
    item: new schema.Union({
      post,
      comment
    }, (value, parent, key) => { // not called
      console.log('schemaAttribute called')
      return parent.itemType
    }),
    createdBy: user,
    forTab: tab
  },
  {
    idAttribute: 'id',
    mergeStrategy: (entityA, entityB) => { // not called
      console.log('mergeStrategy called')
      const merged = entityA.merge(entityB)
      return merged
    },
    processStrategy: entity => {
      const normalized = new RemovalLogRecord(entity)
[...]

PR #265, opened 2017-05-20
added afterStrategy option to schema.Entity

tugorez:
# Problem

**normalizr** options allows you to specify how data will be transformed before being normalized just by passing a **processStrategy** function which is awesome; 

Unfortunately there is no option to post-process  each normalized entity.
One use-case is to cast each entity to an immutable.Record before being handled by a reducer 
https://github.com/paularmstrong/normalizr/issues/206, https://github.com/paularmstrong/normalizr/issues/199,

# Solution

add an **afterStrategy** option (such as **processStrategy**) wich receives each normalized entity and returns the entity transformed.

# TODO

- [x] Add & update tests


paularmstrong:
Please see arguments against this in both linked issues, #199 and #206. I have yet to be convinced that Normalizr needs yet another option, expanded API, and potential for confusion.

PR #269, opened 2017-06-02
Change `result` to `entities`

wrumsby:
# Problem

I'm new to normalizr and playing around trying to `normalize` some data, then `denormalize` it.

Based on what I'm seeing `input` should be `normalize(data).entities` rather than `normalize(data).result`.

# Solution

Modify documentation.


wrumsby:
Looking through the tests it seems the documentation that already exists is correct.

Issue #278, opened 2017-07-21
denormalizing fails when top level id is empty string


dkrieger:
# Problem

Denormalizing fails when the root's id value is an empty string. Setting a nested entity's id value to empty string doesn't cause any issues.

**Input**

Here is my initial data, how I'm normalizing it (and normalization output), and my denormalize call:

```js
const data = {
  slug: "",
  sections: [
    {
      slug: "1",
      sections: [ { slug: "4", sections: [
            { slug: "5", sections: [] }
          ] } ]
    },
    { slug: "2", sections:[] },
    { slug: "3", sections:[] }
  ]
};

const section = new schema.Entity('sections', {}, {idAttribute: "slug"});
const sections = new schema.Array(section);
section.define({ sections });
const menu = new schema.Entity('menu', { sections }, {idAttribute: "slug"});

const normalizedData = normalize(data, menu);
// console.log(JSON.stringify(normalizedData));
// {
//   "entities": {
//     "sections": {
//       "baz": {
//         "slug": "baz",
//         "sections": []
//       },
//       "bar": {
//         "slug": "bar",
//         "sections": [
//           "baz"
//         ]
//       },
//       "foo": {
//         "slug": "foo",
//         "sections": [
//           "bar"
//         ]
//       },
//       "zip": {
//         "slug": "zip",
//         "sections": []
//       },
//       "zap": {
//         "slug": "zap",
//         "sections": []
//       }
//     },
//     "menu": {
//       "": {
//         "slug": "",
//         "sections": [
//           "foo",
//           "zip",
//           "zap"
//         ]
//       }
//     }
//   },
//   "result": ""
// }

const denormalizedNormalizr = JSON.stringify(denormalize(normalizedData.result, menu, normalizedData.entities));
console.log(denormalizedNormalizr);
```

**Output**

Here's what I expect to see when I run the above (same as `data`):

```json
{
  "slug": "",
  "sections": [
    {
      "slug": "foo",
      "sections": [
        {
          "slug": "bar",
          "sections": [
            {
              "slug": "baz",
              "sections": []
            }
          ]
        }
      ]
    },
    {
      "slug": "zip",
      "sections": []
    },
    {
      "slug": "zap",
      "sections": []
    }
  ]
}

```

Here's what I *actually* see when I run the above:

```json
""
```

Note that using denormalizr yields the expected result (using `denormalize(normalizedData.result, normalizedData.entities, menu)`).

dkrieger:
[...]

PR #279, opened 2017-07-21
denormalize: replace falsy input check with undefined input check

dkrieger:
Returning falsy input causes valid ids such as "" and 0 to be returned,
even if they could otherwise be properly denormalized. Loosening this
check to undefined will allow `null` to be processed, for example, but
that shouldn't be too much of an issue.

# Problem

`denormalize` returns unprocessed `input` value when `input` is falsy. `0` and `""` are both falsy but are potentially valid inputs. See https://github.com/paularmstrong/normalizr/issues/278

# Solution

Instead of checking for falsy values, check for undefined. When a normalization has no result, the value given as input will likely be undefined. Some cases that should be and were previously short circuited will  not be (e.g. `false`, `null`), but the value:cost ratio of handling those cases explicitly seems low.

# TODO

- [x] Add & update tests

(note: not sure documentation needs to be updated here, as this is fixing a bug)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -117,9 +117,7 @@ const getEntities = (entities) => {
 };
 
 export const denormalize = (input, schema, entities) => {
-  if (!input) {
-    return input;
+  if (typeof input !== 'undefined') {
+    return getUnvisit(entities)(input, schema);
   }
-
-  return getUnvisit(entities)(input, schema);
 };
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -117,9 +117,7 @@ const getEntities = (entities) => {
 };
 
 export const denormalize = (input, schema, entities) => {
-  if (!input) {
-    return input;
+  if (typeof input !== 'undefined') {
+    return getUnvisit(entities)(input, schema);
```
dkrieger: if input is undefined, no need to return input -- not returning anything implicitly returns undefined.

PR #283, opened 2017-07-31
Introduce `Map` schema

praxxis:
# Problem

GraphQL does [not support maps](https://github.com/facebook/graphql/issues/101). Instead, users are requied to use an array of objects to represent nested data.

Redux (and Mobx) users, on the other hand, are [encouraged](http://redux.js.org/docs/recipes/reducers/NormalizingStateShape.html) to store their data in maps. 

Currently normalizr does not provide an easy way to convert between the two paradigms. 

# Solution

I've added a new Schema based on the Values schema that transforms an array-of-objects into a map.

*Example*

```js
const data = [ { id: 1, name: 'oliver' }, { id: 2, name: 'waldo' } ];

const pet = new schema.Entity('pets');
const mapSchema = new schema.Map(pet);

const normalizedData = normalize(data, mapSchema);
```

*Output*

```js
{
  entities: {
    pets: { '1': { id: 1, name: 'oliver' }, '2': { id: 2, name: 'waldo' } }
  },
  result: { '1': 1, '2': 2 }
}
```

# TODO

- [x] Add & update tests

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -8,6 +8,7 @@
   - [Object](#objectdefinition)
   - [Union](#uniondefinition-schemaattribute)
   - [Values](#valuesdefinition-schemaattribute)
+  - [Map](#mapdefinition-schemaattribute)
 
 ## `normalize(data, schema)`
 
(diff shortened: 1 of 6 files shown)


paularmstrong:
This is exactly what `[ pet ]` or `new schema.Array(pet)` does. How does `Map` handle things differently?

praxxis:
Given the following:
```js
const students = new schema.Entity('students');
const classrooms = new schema.Entity('classrooms', {
  students: [students]
});

const classroom = {
  id: "1",
  students: [{id: "1", name: "bob"}, {id: "2", name: "fred"}]
};

const result = normalizr.normalize(classroom, classrooms);
```

you get

```js
{
    "entities": {
        "students": {
            "1": {
                "id": "1",
                "name": "bob"
            },
            "2": {
                "id": "2",
                "name": "fred"
            }
        },
        "classrooms": {
            "1": {
                "id": "1",
                "students": ["1", "2"]
            }
        }
    },
    "result": "1"
}
```

Note that `entities.classrooms.students` is an array. With `schema.Map`, `students` would be `{1: "1", 2: "2"}`.

Granted, this level of transformation of an entity might not be desired, but the GraphQL array-of-objects / map-of-objects transform seems like it'd be pretty common, given the documentation for both projects. If you don't feel this PR is appropriate that's fine - I figured I'd put up a solution in case anyone else has similar needs to me :)

paularmstrong:
I think this is over-complicating the things that Normalizr is able to do. It's already pretty hard to document what it means to do `schema.Values` vs `schema.Array` vs `schema.Union`. I fear that another option complicates things a little too much.

That being said, you can absolutely write your own schema class (just copy out schema/Map.js to your own project), extend it from `schema.Values` and use it anywhere. I don't plan on dropping support for extending your own schema types. 

I'll keep this open for a bit, hear more arguments if there are any, and think more on whether the additional option is worth the potential for confusion.

praxxis:
Not a problem. Assuming this isn't merged, would you be adverse to a PR that extends the TypeScript definition? At the moment I have to use a few `(this as any)`'s because TS doesn't know about the protected/private methods.

paularmstrong:
> would you be adverse to a PR that extends the TypeScript definition? 

I've never used TypeScript and don't foresee that going into my future. All the help there that you can provide is appreciated!

PR #294, opened 2017-09-28
Use input as id for unvisitEntity in denormalize

nikita-ushakov:
# Problem

Denormalizing fails if `idAttribute` option of `EntitySchema` used as a function which uses `parent` or `key` parameters.

Consider this use case

```
const data = [
    { id: '1', guest: null, name: 'Esther' },
    { id: '2', guest: {guest_id: 1}, name: 'Tom' }
]

const guestSchema = new schema.Entity('guests', {}, {
    idAttribute: (value, parent, key) => `${key}-${parent.id}-${value.guest_id}`
})

const patronsSchema = new schema.Entity('patrons', {
    guest: guestSchema
})
const normalizedData = normalize(data, [patronsSchema])
const denormalizedData = denormalize(
    normalizedData.result, 
    [patronsSchema], 
    normalizedData.entities
)
```
Expected `denormalizedData` to be equal to `data` but got error

`TypeError: Cannot read property 'id' of undefined`

# Solution

`unvisitEntity` calls `schema.getId(entity);` which is useless because 1)not all nesessary parameters passed and 2) we already have `input` as id

# TODO

- [x] Add & update tests

### DIFF
--- a/src/__tests__/index.test.js
+++ b/src/__tests__/index.test.js
@@ -224,4 +224,27 @@ describe('denormalize', () => {
     });
     expect(() => denormalize('123', article, entities)).not.toThrow();
   });
+
+  it('denormalizes with function as idAttribute', () => {
+    const normalizedData = {
+      'entities': {
+        'patrons': {
+          '1': { 'id': '1', 'guest': null, 'name': 'Esther' },
+          '2': { 'id': '2', 'guest': 'guest-2-1', 'name': 'Tom' }
+        },
+        'guests': { 'guest-2-1': { 'guest_id': 1 } }
+      },
+      'result': [ '1', '2' ]
+    };
+
+    const guestSchema = new schema.Entity('guests', {}, {
+      idAttribute: (value, parent, key) => `${key}-${parent.id}-${value.guest_id}`
+    });
+
+    const patronsSchema = new schema.Entity('patrons', {
+      guest: guestSchema
+    });
+
+    expect(denormalize(normalizedData.result, [ patronsSchema ], normalizedData.entities)).toMatchSnapshot();
+  });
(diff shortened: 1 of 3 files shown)


src/__tests__/index.test.js
```diff
@@ -165,6 +165,8 @@ describe('denormalize', () => {
     const user = new schema.Entity('users');
     const comment = new schema.Entity('comments', {
       user: user
+    }, {
+      idAttribute: (value, parent, key) => `${value.id}-${parent.id}-${key}`
```
paularmstrong: Can we make a separate test case that's specific to this?

nikita-ushakov:
- [x] Added saparete test for this case

src/__tests__/index.test.js
nikita-ushakov: Sure. Done

vicmmontqb:
This is exactly an error I ran into yesterday! Please check in the fix! Thanks for looking at this issue!

nikita-ushakov:
Is there any issue with this PR? Will it be merged?
Thank you @paularmstrong for reviewing my PR and maintaining this package in general

paularmstrong:
@nikita-ushakov sorry nothing wrong, just been busy

paularmstrong:
Available in v3.2.4

vicmmontqb:
Thank you!

diegohaz:
I think this change introduced a regression. My tests on https://github.com/diegohaz/arc/tree/redux were passing with v3.2.3, but started to fail with v3.2.4.

I've cloned this repo and tested my project against every commit after v3.2.3 and it has failed exactly here. Will try to investigate it.

Issue #296, opened 2017-10-03
Denormalizing circular references 🔄


PhilippSpo:
I am a bit confused about how this should work. The docs say:
> If your schema and data have recursive references, only the first instance of an entity will be given. Subsequent references will be returned as the id provided.

Given this Schema and Entities:
```js
const user = new schema.Entity('users');
const report = new schema.Entity('reports');

user.define({
  reports: [report]
});
report.define({
  draftedBy: user,
});

const entities = {
  reports: {
    '1': {
      id: '1',
      draftedBy: '2',
    }
  },
  users: {
    '2': {
      id: '2',
      reports: ['1']
    }
  }
};
```
I would expect the following output when denormalizing
```js
expect(denormalize('1', report, entities)).toEqual({
  id: '1',
  draftedBy: {
    id: '2',
    reports: ['1']
  }
})
```

However what we get when calling denormalize is a circular structure:
![image](https://user-images.githubusercontent.com/8394418/31134969-f9ae2f38-a863-11e7-96f8-fea36c1b3a25.png)

I think [this test](https://github.com/paularmstrong/normalizr/blob/8dbcd8d440855b46defecd7b949f6298b924d378/src/schemas/__tests__/Entity.test.js#L210) is supposed to test the behavior I just described.

Looking at the snapshot of the non-immutable JS version we can also see a Circular reference:
https://github.com/paularmstrong/normalizr/blob/c9c8c8bfd616c76446c3d687c6c6d3cd1c48550c/src/schemas/__tests__/__snapshots__/Entity.test.js.snap#L98-L116

So it seems like the implementation is not following the documentation here.

I also created a [CodeSandbox](https://codesandbox.io/s/rl4ll6vx8m) for this issue. 👀

onoufriosm:
Any news on this? Is there a way to get a denormalized output without circular structure? As of v3.2.4 I still get recursive references in my denormalized output.

PhilippSpo:
@onoufriosm having circular dependencies is not a bug, but a feature. So it's not the implementation not following the docs but the other way around 😉. Having circular reference is indeed pretty handy because there are no dead ends.

However, there is one problem: You can't `normalize` an entity that was produced by `denormalizing` it from e.g. the store. This crashes because the `normalize` function can't handle circular references.

sarink:
What if all relationships were defined as `getters`?

paularmstrong:
This is as-designed.

PR #297, opened 2017-10-05
Avoid circular structures when denormalizing

PhilippSpo:
### Problem

The problem is outlined in detail in [this issue](https://github.com/paularmstrong/normalizr/issues/296) and can be clearly seen in the snapshot tests:
https://github.com/paularmstrong/normalizr/blob/c9c8c8bfd616c76446c3d687c6c6d3cd1c48550c/src/schemas/__tests__/__snapshots__/Entity.test.js.snap#L98-L116

#### Changing the denormalization algorithm to always pass down the list of ancestors.
When denormalizing an `Entity` we pass all already expanded references of the subtree as the `cache` parameter.

This has two benefits:
- find out whether or not to denormalize a given entity by looking into the `cache`
- avoids #233 by default

### TODO

- [x] Add & update tests

### DIFF
--- a/src/schemas/__tests__/__snapshots__/Entity.test.js.snap
+++ b/src/schemas/__tests__/__snapshots__/Entity.test.js.snap
@@ -99,7 +99,7 @@ Object {
   "draftedBy": Object {
     "id": "456",
     "reports": Array [
-      [Circular],
+      "123",
     ],
     "role": "manager",
   },
@@ -107,7 +107,7 @@ Object {
   "publishedBy": Object {
     "id": "456",
     "reports": Array [
-      [Circular],
+      "123",
     ],
     "role": "manager",
   },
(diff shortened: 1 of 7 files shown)


src/schemas/__tests__/__snapshots__/Entity.test.js.snap
```diff
@@ -94,93 +94,71 @@ Object {
...
 `;
 
-exports[`EntitySchema denormalization denormalizes recursive dependencies 1`] = `
+exports[`EntitySchema denormalization denormalizes recursive dependencies (immutable) 1`] = `
 Object {
   "draftedBy": Object {
     "id": "456",
     "reports": Array [
-      [Circular],
+      "123",
```
PhilippSpo: No more circular structure 🎉

paularmstrong:
This is a desired feature. See #239

PhilippSpo:
Interesting 🤔 That means you can't call `normalize` on the result of `denormalize` if the entity contains recursion?

PhilippSpo:
I can definitely understand the use case for allowing circular reference when denormalizing. But as mentioned above you can't normalize an entity that contains a circular reference, which is a problem in my use case.

I could apply the above described caching strategy to `normalize` enabling it to handle circular references by avoiding to visit an entity it has already been visited in the current subtree. WDYT @paularmstrong ?

Issue #200, opened 2017-01-02
Wanted: Flux example


paularmstrong:
# Problem

A simple and easy to follow example of using Normalizr with Flux would be a great addition to the [`/examples`](https://github.com/paularmstrong/normalizr/tree/master/examples) available.

# Solution

I haven't used Flux for over a year, so am not a good person to go and try implementing any example using it. This is a good chance for anyone looking to give back to open source to get started!

isikfsc:
It would be great if this includes **many-to-many** relationships besides the more commonly exampled **one-to-many** relationships.

wuweiweiwu:
@paularmstrong I can take this is no one has yet :)

paularmstrong:
@wuweiweiwu Still open!

wuweiweiwu:
@paularmstrong on it!

konoufo:
This can be closed now right ?

paularmstrong:
No, the PR was closed by the contributor before it was able to be merged.

paularmstrong:
Seems like no one uses flux `¯\_(ツ)_/¯`

PR #317, opened 2018-01-21
Adding Flux example

wuweiweiwu:
# Problem

Flux example usage with normalizr

#200 

# Solution

An example using Flux using the Github API. Implemented issues and commits.


wuweiweiwu:
Build failed because rimraf command not found. Not sure why

PR #318, opened 2018-01-22
Refactored build process, distributing module entry

Andarist:
# Problem

No "module" entry. Having this allows bundlers to apply optimization techniques such as tree-shaking and scope hoisting.

# Solution

I've updated several related deps and added a new output entry in rollup.config.js

If you have any questions why certain changes were made, please just ask and I will provide any necessary information.

# TODO

- [x] Add & update tests
- [x] Ensure CI is passing (lint, tests, flow)
- [ ] Update relevant documentation

PR #323, opened 2018-02-08
Deduplicated babel config

Andarist:
Follow up to https://github.com/paularmstrong/normalizr/pull/318#discussion_r166977224

# TODO

- [ ] Add & update tests

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -1,0 +1,13 @@
+const { NODE_ENV, BABEL_ENV } = process.env;
+
+const cjs = BABEL_ENV === 'cjs' || NODE_ENV === 'test';
+
+module.exports = {
+  presets: [['env', { modules: false, loose: true }]],
+  plugins: [
+    cjs && 'transform-es2015-modules-commonjs',
+    'transform-object-rest-spread',
+    // TODO: use 'loose' mode for this after upgrading to babel@7
+    'transform-class-properties'
+  ].filter(Boolean)
+};
--- a/package.json
+++ b/package.json
@@ -10,20 +10,8 @@
     "url": "https://github.com/paularmstrong/normalizr.git",
     "type": "git"
   },
-  "keywords": [
-    "flux",
-    "redux",
-    "normalize",
-    "denormalize",
-    "api",
-    "json"
-  ],
-  "files": [
-    "dist/",
-    "index.d.ts",
-    "LICENSE",
-    "README.md"
-  ],
+  "keywords": ["flux", "redux", "normalize", "denormalize", "api", "json"],
+  "files": ["dist/", "index.d.ts", "LICENSE", "README.md"],
   "main": "dist/normalizr.js",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -10,20 +10,8 @@
...
-    "api",
-    "json"
-  ],
-  "files": [
-    "dist/",
-    "index.d.ts",
-    "LICENSE",
-    "README.md"
-  ],
+  "keywords": ["flux", "redux", "normalize", "denormalize", "api", "json"],
```
Andarist: those changes are caused by some auto-formatting - im not sure, but I think my yarn does it automatically when installing new deps (which I did for a brief moment, but removed later)

I can revert this file if you want

rollup.config.js
```diff
@@ -20,9 +20,7 @@ export default {
   ],
   plugins: [
     babel({
-      babelrc: false,
-      presets: [ [ 'env', { modules: false, loose: true } ] ],
-      plugins: [ 'external-helpers', 'babel-plugin-transform-object-rest-spread' ]
+      plugins: ['external-helpers']
```
Andarist: this is gonna be merged automatically with root babel config

.babelrc.js
```diff
@@ -0,0 +1,13 @@
+const { NODE_ENV, BABEL_ENV } = process.env;
+
+const cjs = BABEL_ENV === 'cjs' || NODE_ENV === 'test';
+
+module.exports = {
+  presets: [['env', { modules: false, loose: true }]],
+  plugins: [
+    cjs && 'transform-es2015-modules-commonjs',
```
Andarist: i dont use `loose` for cjs, because it "breaks" importing namespaces - it uses a simple assignment for `__esModule` property thus making it enumerable, and spreading an imported namespace onto some object carries over `__esModule` - gain of using loose in this case is minimal and I tend to avoid it for this case

.babelrc.js
```diff
@@ -0,0 +1,13 @@
+const { NODE_ENV, BABEL_ENV } = process.env;
+
+const cjs = BABEL_ENV === 'cjs' || NODE_ENV === 'test';
```
Andarist: this makes cjs transform an opt-in behaviour, rather an opt out - which is believe is safer choice, because if you dont opt-in into it your tools etc will fail fast and that will have you to explicitly opt-in into that transform

when you have to opt out of it it makes it easy to end up with deoptimized builds etc if a tool (i.e. webpack) knows how to handle both esm and cjs (in this situation it would gladly accept cjs input without any warnings that ESM format can be used)

package.json
paularmstrong: I'm in the process of adding prettier auto-formatting on pre-commit :) Don't worry about it.

.babelrc.js
paularmstrong: I actually just added the bottom two plugins, but missed the commonjs. I'll get this all merged on my end

PR #328, opened 2018-03-09
Add example how to rename a field

iamandrewluca:


paularmstrong:
Thanks, but is this really necessary? I would prefer if documentation were as short as possible to reduce extra overhead when skimming through.

iamandrewluca:
@paularmstrong I agree with you. Made the PR because for me was a bit harder to realise how to do this, and thought to contribute. At your desire you can close this PR no problems 😃

PR #330, opened 2018-04-06
Add support for circular references in JS objects

jeffreyyoung:
# Problem

Unable to normalize objects with circular references.  I get a `maximum call stack size exceeded` error. In my opinion since already normalizr supports denormalizing circular references, it makes sense to support normalizing circular references.

# Solution
If we have already normalized an entity, don't normalize it again in order to avoid infinite normalization loops.

I pass a `visitedEntities` set into all normalize functions.  In the entity normalize method I added the following logic: 
If an entity has already been normalized, don't normalize it and just return the id, 
else normalize the enitity and add the entity to the visitedEntities set

# notes
figure out why npm test works but npm run precommit brakes when I uncomment the second added test

# TODO
- [ ] Add & update tests

### DIFF
--- a/src/__tests__/index.test.js
+++ b/src/__tests__/index.test.js
@@ -18,6 +18,43 @@ describe('normalize', () => {
     expect(normalize([{ id: 1, type: 'foo' }, { id: 2, type: 'bar' }], [mySchema])).toMatchSnapshot();
   });
 
+  it('normalize entities with circular references', () => {
+    const user = new schema.Entity('users');
+    user.define({
+      friends: [user]
+    });
+
+    const myUser = { id: 234, friends: [] };
+    myUser.friends.push(myUser);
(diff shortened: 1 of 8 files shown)


jeffreyyoung:
ahh that makes sense.  Is this a feature that you'd want to add to normalizr?  If so I can fix the tests to use snapshot and remove the `Set` from the code

paularmstrong:
@jeffreyyoung yeah, seems useful and covers the bases without being a bloaty. Happy to merge it with the couple fixes if you get a chance

jeffreyyoung:
@paularmstrong sounds good.  I'll get it in sometime this weekend 👍

paularmstrong:
@jeffreyyoung Thanks for taking some stabs at this.  I'm going to go with the solution in #335

PR #334, opened 2018-05-10
feat: Adds ability to define custom Class for entities objects

Todmy:
# Problem

Inability to use a custom class for entities

# Solution

Hi guys.
Thank you for this awesome tool! 
Recently I have faced the need to split and simplify the logic of our project. And I thought it will be awesome to use custom classes for entities.
What is my motivation? As I said, to move the low-level logic of interaction inside entities to the separate class. Also, it will help to use some custom methods for adding new instances to entities, e.g. `normalizedData.entities.tweets.$customAdd(newTweet)`
For clarity, you can check simple example here: https://codesandbox.io/s/50m0n9mk5n (src/index.js)
What do you think about this idea?

p.s.: If you are OK with it, I can update docs and tests.

# TODO

- [ ] Add & update tests

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -22,7 +22,7 @@ const addEntities = (entities) => (schema, processedEntity, value, parent, key) 
   const schemaKey = schema.key;
   const id = schema.getId(value, parent, key);
   if (!(schemaKey in entities)) {
-    entities[schemaKey] = {};
+    entities[schemaKey] = schema.constructEntity();
   }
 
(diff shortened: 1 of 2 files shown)


paularmstrong:
I think this is neat! However, I'd rather it live in "user-land", since it seems like a pretty specific need that normalizr doesn't need to solve at this time.

Todmy:
Hi, but why? you know that this will not break functionality and leave normalizr simple as it is. And just extends library for advanced usage.

PR #357, opened 2018-10-09
added a shared context object

gtg092x:
# Problem

I need a shared state between processStrategy calls to check for circular dependencies.

# Solution

Adding an optional fourth parameter (context) gives me access to a shared object I can use to add arrays of seen objects, etc. The context object also contains the current accumulating entities object by default.

# TODO

- [x] Add & update tests

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -3,11 +3,11 @@
 * [normalize](#normalizedata-schema)
 * [denormalize](#denormalizeinput-schema-entities)
 * [schema](#schema)
-  - [Array](#arraydefinition-schemaattribute)
-  - [Entity](#entitykey-definition---options--)
-  - [Object](#objectdefinition)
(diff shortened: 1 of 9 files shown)


paularmstrong:
I don't really understand the problem you're trying to solve. Maybe the test could show a more real-world use-case?

Are you maybe looking for `mergeStrategy` to handle when you're receiving duplicate items? https://github.com/paularmstrong/normalizr/blob/master/docs/api.md#entitykey-definition---options--

gtg092x:
Thanks for responding @paularmstrong ! As far as I could tell, there wasn't a way to use a merge strategy for circular dependencies. It might help if I put together a more thorough solution explaining the problem and how this could solve it. 

I'll put one together in examples and follow up.

paularmstrong:
@gtg092x I'm doing some cleanup, please reopen if you're still working on this.

PR #365, opened 2019-02-24
[minor] Improve typescript types with generics

ntucker:
# Problem
It's very hard to infer anything about the types without generics. Given that it's most useful to have types for the data stored in your normalized cache this seems pretty essential.

# Solution
The idea here is at some point you'll point to an entity whose processStrategy will return the type you expect.

Splitting the Schema union type was essential to distinguish between list results and singular results as well.

In the end I was able to create my own type in userland that could extract what type to expect:
```typescript
export type SchemaOf<T> = T extends SchemaArray<infer R>
  ? R[]
  : T extends SchemaBase<infer R>
    ? R
    : never;
```

# TODO

- [x] Add & update tests

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,59 +1,62 @@
 declare namespace schema {
-  export type StrategyFunction = (value: any, parent: any, key: string) => any;
+  export type StrategyFunction<T> = (value: any, parent: any, key: string) => T;
   export type SchemaFunction = (value: any, parent: any, key: string) => string;
   export type MergeFunction = (entityA: any, entityB: any) => any;
 
-  export class Array {
-    constructor(definition: Schema, schemaAttribute?: string | SchemaFunction)
+  export class Array<T = any> {
+    constructor(definition: Schema<T>, schemaAttribute?: string | SchemaFunction)
     define(definition: Schema): void
   }
 
-  export interface EntityOptions {
+  export interface EntityOptions<T = any> {
     idAttribute?: string | SchemaFunction
     mergeStrategy?: MergeFunction
-    processStrategy?: StrategyFunction
+    processStrategy?: StrategyFunction<T>
   }
 
-  export class Entity {
-    constructor(key: string | symbol, definition?: Schema, options?: EntityOptions)
+  export class Entity<T = any> {
+    constructor(key: string | symbol, definition?: Schema, options?: EntityOptions<T>)
     define(definition: Schema): void
     key: string
+    getId: SchemaFunction
+    _processStrategy: StrategyFunction<T>
   }
 
-  export class Object {
-    constructor(definition: {[key: string]: Schema})
+  export class Object<T = any> {
+    constructor(definition: {[key: string]: Schema<T>})
     define(definition: Schema): void
   }
 
-  export class Union {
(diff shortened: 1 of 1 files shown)


paularmstrong:
@ntucker can you rebase and get tests passing?

paularmstrong:
ping @ntucker I'd like to get a new release out soon and include this if possible!

ntucker:
Yeah, I'll take a look. Been super busy with end-of-quarter crunch time.

paularmstrong:
> Yeah, I'll take a look. Been super busy with end-of-quarter crunch time.

I know that feeling! If you can't get to it, that's fine and I'll cut a release now. Just let me know!

ntucker:
rebased, and I'll update the tests next

ntucker:
actually may do a bit of changes here so don't merge yet

ntucker:
update i'll need to redo some of the types a little bit. i'll do it this weekend.

paularmstrong:
Sounds good @ntucker!

paularmstrong:
ping @ntucker any update ready?

ntucker:
@paularmstrong  sorry about the huge delay. Got distracted by other stuff and I hadn't quite figured out how to completely do this right. Did some research today and got it all done nicely tho :)

iscriptology:
Any plan to merge?

ntucker:
I think he checks this project once every 3 months or so. @paularmstrong do you need more maintainers? I would happily help.

paularmstrong:
@ntucker I check often, but get side-tracked with higher priorities and forget to come back. I'd be happy to make you a maintainer—you've been really helpful!

paularmstrong:
Available in v3.4.1

lauritzsh:
This breaks our build unfortunately (I have to type the `E` [here](https://github.com/paularmstrong/normalizr/pull/365/files#diff-b52768974e6bc0faccb7d4b75b162c99R57)). Should this be a major version bump instead?

ntucker:
what type do you put for E?

lauritzsh:
The object of denormalized entities as in `{entities: E, result: R}`. I didn't have to do that before since it defaulted to `any`. It happens because we normalize our API calls and we type the response data (something like 

```typescript
const normalized = normalize(data, schema);
const response: Success = { ...normalized, error: false };

return response;
```

Now the response errors since it expect some specific keys but instead just get `{ [key: string]: any }`. I ended up sort of doing `normalize<unknown, Entities, string>` where `Entities` describes the expected denormalized entities of the API call.

ntucker:
@lauritzsh would it be better if T was the last argument of the generic?

```typescript
export function normalize<E = { [key: string]: { [key: string]: T } }, R = any, T = any>(
  data: any,
  schema: Schema<T>
): NormalizedSchema<E, R>;
```

then you wouldn't have to put the default. Otherwise right now the ordering is definitely backward breaking.

lauritzsh:
That was my thought as well but I don't think that's possible due to `Type parameter defaults can only reference previously declared type parameters.` At least I get that error when I try. Not sure if this can be done without breaking the API and then it should probably be a major version bump instead of patch.

ntucker:
Ah right; no wonder I did it that way :p

@paularmstrong maybe this needs a different kind of version bump due to breaking type definitions.

PR #374, opened 2019-05-15
Added BaseSchema for typesafe schema walking

jonaswre:
# Problem
If you try to iterate through Schemas in other software projects you end up doing `(schema as any).schema` because its not defined in the class.

# Solution
Add schema to the class

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3,65 +3,58 @@ declare namespace schema {
   export type SchemaFunction = (value: any, parent: any, key: string) => string;
   export type MergeFunction = (entityA: any, entityB: any) => any;
 
-  export class Array extends BaseSchema {
-    constructor(definition: Schema, schemaAttribute?: string | SchemaFunction)
-    define(definition: Schema): void
+  export class Array implements BaseSchema {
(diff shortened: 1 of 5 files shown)


paularmstrong:
> If you try to iterate through Schemas

Can you help me understand what this means and why you're doing it? An example use-case with code would really help.

jonaswre:
I use your package within another package. I want to find all child Entites but i cant iterate through the child schemas because the schema property is not include in the type definition. Peusdo code looks like this:
`function getChildEntites(entitySchema: schema.Entity) {
   Object.entites((entitySchema as any).schema).foreach(([key, schema], [string, Schema]) => {....]);
}`
`(entitySchema as any).schema` feels very dirty to do an any cast just because its missing in the typedefinition.

paularmstrong:
> feels very dirty to do an any cast just because its missing in the typedefinition.

You should be doing a union type to check against the possible types. eg `schema.Entity | schema.Array | ...`. I'm going to close this in favor of that, since that seems safe enough. Please reopen if you are experiencing further issues.

Your PR also doesn't account for a schema of type `{}` or `[]`, which are also valid.

PR #380, opened 2019-06-05
Fix schema key is not in entities

ktmouk:
# Problem

Currently, `denormalize` sets value to `undefined` If did not find entity.

Example:
```javascript
const entities = {
  comments: { '1': { user: '123' } },
  users: {}  // 123 is not found
};
console.log(denormalize('1', comment, entities));  // { user: undefined }
```

But, If entities don't have "schema key", `denormalize` throw exception.

Example:
```javascript
const entities = {
  comments: { '1': { user: '123' } },
  // user: {}
};
console.log(denormalize('1', comment, entities));  // TypeError: Cannot read property '123' of undefined
```

I think the correct behavior is to set undefined if the entities have no schema key.

# Solution
Add `entities[schemaKey]` check to `getEntities` method.

# TODO

- [x] Add & update tests


paularmstrong:
Available in v3.4.1

PR #387, opened 2019-06-20
fix(schemas): don't use schema to attribute mapping on singular array schemas

fpipita:
# Problem

Denormalization does not seem to produce the correct result when a singular Array schema owns an Object schema whose instances own an `id` property. 

For example:

```javascript
const cats = new schema.Entity('cats');
// catRecord instances own an `id` property
const catRecord = new schema.Object({
   cat: cats
});
const catList = new schema.Array(catRecord);
```

The problem is in the the `denormalizeValue` method of the `Polymorphic` class.

When the value to be unvisited owns a property named `id`, then this method uses the value of the `id` property as the input value for the call to `unvisit` in place of the value itself, which is not the case on singular Array schema, where the value itself should always be used instead.

# Solution

I've added a guard which prevents the value of the `id` property to become the input to `unvisit` when the current schema is a singular schema.

# TODO

- [ x] Add & update tests
- [ x] Ensure CI is passing (lint, tests, flow)

### DIFF
--- a/src/schemas/Polymorphic.js
+++ b/src/schemas/Polymorphic.js
@@ -45,7 +45,7 @@ export default class PolymorphicSchema {
     if (!this.isSingleSchema && !schemaKey) {
       return value;
     }
-    const id = isImmutable(value) ? value.get('id') : value.id;
+    const id = this.isSingleSchema ? undefined : isImmutable(value) ? value.get('id') : value.id;
     const schema = this.isSingleSchema ? this.schema : this.schema[schemaKey];
     return unvisit(id || value, schema);
   }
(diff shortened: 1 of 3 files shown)


src/schemas/Polymorphic.js
```diff
@@ -45,7 +45,7 @@ export default class PolymorphicSchema {
     if (!this.isSingleSchema && !schemaKey) {
       return value;
     }
-    const id = isImmutable(value) ? value.get('id') : value.id;
+    const id = this.isSingleSchema ? void 0 : isImmutable(value) ? value.get('id') : value.id;
```
paularmstrong: To avoid confusion for others, since `void 0` isn't common, can we use a canonical `undefined` here?

fpipita: Sure, I've just updated the patch and rebased it to upstream/master.

PR #397, opened 2019-07-12
[enhance] Allow schemas to have extraneous attributes that are simply ignored

ntucker:
# Problem

Sometimes it might be useful to express more in a schema object to be used by other utilities. For instance, having default values for other attributes that are expected in a response but don't turn into entities.

# Solution

Pretty simple: in unvisit we check if schema is really a schema, otherwise we just return input.

# TODO

- [x] Add & update tests

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -95,7 +95,11 @@ const getUnvisit = (entities) => {
       return unvisitEntity(input, schema, unvisit, getEntity, cache);
     }
 
-    return schema.denormalize(input, unvisit);
+    if (typeof schema.denormalize === 'function') {
+      return schema.denormalize(input, unvisit);
+    }
+
+    return input;
   };
 };
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,38 +4,42 @@ declare namespace schema {
   export type MergeFunction = (entityA: any, entityB: any) => any;
 
   export class Array<T = any> {
-    constructor(definition: Schema<T>, schemaAttribute?: string | SchemaFunction)
-    define(definition: Schema): void
+    constructor(definition: Schema<T>, schemaAttribute?: string | SchemaFunction);
+    define(definition: Schema): void;
   }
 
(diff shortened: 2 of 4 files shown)


ntucker:
Rebased this on master since the other PR landed.

src/index.js
```diff
@@ -95,7 +95,11 @@ const getUnvisit = (entities) => {
       return unvisitEntity(input, schema, unvisit, getEntity, cache);
     }
 
-    return schema.denormalize(input, unvisit);
+    if (typeof schema.denormalize === 'function') {
```
paularmstrong: in what case does this function not exist? All schema types should have a `denormalize` function

ntucker: when typeof schema === 'string' | 'number' | 'symbol'

ntucker: perhaps I should remove lines 90-92 to make the primitive types be consistent? I feel the fallback is more general because there's probably typeofs we can't even anticipate - like symbol didn't used to exist.

ntucker: @paularmstrong ping

paularmstrong:
This is a really confusing addition. I'm not sure I understand the use case, because normalizr already allows pass-through of unknown data. 

One of your tests uses this:
```js
normalize(
  { data: [{ id: 1, type: 'foo' }, { id: 2, type: 'bar' }], extra: 'five' },
  { data: [mySchema], extra: '' }
)
```

But if you normalize it a little more simply:
```js
normalize(
  { data: [{ id: 1, type: 'foo' }, { id: 2, type: 'bar' }], extra: 'five' },
  { data: [mySchema] }
);
```

The result should the same:
```js
{
  entities: { tacos: { 1: {...}, 2: {...} } },
  result: { data: [1, 2], extra: 'five' }
}
```

> Sometimes it might be useful to express more in a schema object to be used by other utilities. 

I don't want normalizr's schema to be depended on by other packages. That's too much overhead and care needed to ensure compatibility.

PR #411, opened 2019-08-06
[fix] When normalize() receives null input, don't say it is an object

ntucker:
# Problem
In Javascript typeof null === 'object'. 

# Solution

Since the check in normalize includes one for !input, it should say 'found null' when null is passed rather than a confusing message that it found object when it needed object.

# TODO

- [x] Add & update tests

### DIFF
--- a/src/__tests__/index.test.js
+++ b/src/__tests__/index.test.js
@@ -12,6 +12,11 @@ describe('normalize', () => {
     expect(() => normalize({})).toThrow();
   });
 
+  test('cannot normalize with null input', () => {
+    const mySchema = new schema.Entity('tacos');
+    expect(() => normalize(null, mySchema)).toThrow(/null/);
+  });
+
   test('normalizes entities', () => {
     const mySchema = new schema.Entity('tacos');
 
--- a/src/index.js
+++ b/src/index.js
@@ -44,7 +44,9 @@ export const schema = {
 export const normalize = (input, schema) => {
   if (!input || typeof input !== 'object') {
     throw new Error(
-      `Unexpected input given to normalize. Expected type to be "object", found "${input && typeof input}".`
+      `Unexpected input given to normalize. Expected type to be "object", found "${
+        input === null ? 'null' : typeof input
+      }".`
     );
   }
 
--- a/src/index.js
+++ b/src/index.js
@@ -43,7 +43,9 @@ export const schema = {
 
 export const normalize = (input, schema) => {
   if (!input || typeof input !== 'object') {
-    throw new Error(`Unexpected input given to normalize. Expected type to be "object", found "${typeof input}".`);
+    throw new Error(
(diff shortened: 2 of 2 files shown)


src/__tests__/index.test.js
```diff
@@ -12,6 +12,11 @@ describe('normalize', () => {
     expect(() => normalize({})).toThrow();
   });
 
+  test('cannot normalize with null input', () => {
+    const mySchema = new schema.Entity('tacos');
+    expect(() => normalize(null, mySchema)).toThrow(/null/);
```
paularmstrong: nit: `toThrow('Unexpected input given to normalize')`

src/index.js
```diff
@@ -43,7 +43,9 @@ export const schema = {
 
 export const normalize = (input, schema) => {
   if (!input || typeof input !== 'object') {
-    throw new Error(`Unexpected input given to normalize. Expected type to be "object", found "${typeof input}".`);
+    throw new Error(
+      `Unexpected input given to normalize. Expected type to be "object", found "${input && typeof input}".`
```
paularmstrong: Maybe derive the type from the schema? `array` if  `Array.isArray(schema) || schema instanceof schema.Array`

ntucker: I don't believe an array would trigger this. typeof myarray === 'object' and all arrays are truthy.

src/__tests__/index.test.js
ntucker: Just making sure it has null in the error string. Not trying to worry about the rest of the wording since there isn't an conditional in there.

src/index.js
ntucker: null, undefined, numbers, strings, symbols, functions (and thus static classes as well) are the only types that would trigger this. maybe some other primitive type I'm not aware of.

ntucker:
Realized it would be weird to output other falsy things like `0`, so I explicitly only handle null case now.

src/index.js
paularmstrong: I mean derive the type for `Expected type to be "${Array.isArray(schema) || schema instanceof schema.Array ? 'array' : 'object'}"`

ntucker: ahh, gotcha. yes this is a good additional improvement

ntucker: but then shouldn't it fail if you get an array when expecting a plain object? currently this check will pass

ntucker: Should I add a respective test for array vs plain object in the if statement?

ntucker: Seems like array and objects are interchangable due to tests like:

```js
    test('normalizes Objects using their values', () => {
      const userSchema = new schema.Entity('user');
      const users = new schema.Array(userSchema);
      expect(normalize({ foo: { id: 1 }, bar: { id: 2 } }, users)).toMatchSnapshot();
    });
```

PR #416, opened 2019-08-27
Correct usage of word "entities" in documentation

OliverJAsh:


docs/api.md
```diff
@@ -76,7 +76,7 @@ const denormalizedData = denormalize({ users: [1, 2] }, mySchema, entities);
 
 ### `Array(definition, schemaAttribute)`
 
-Creates a schema to normalize an array of entities. If the input value is an `Object` instead of an `Array`, the normalized result will be an `Array` of the `Object`'s values.
```
OliverJAsh: I initially thought `Array` only worked with `Entity` schemas, but it works with all schema types, e.g. `new schema.Array(myObjectSchema)`.

PR #419, opened 2019-09-03
[enhance] Use @babel/runtime with es builds to reduce bundle size

ntucker:
# Problem
Applications using webpack or other libraries using rollup can't take full advantage of es modules when the babel runtime helpers are just copied to every file. Including this doesn't change a thing about the non-es bundles but potentially saves great bundle sizes.

# Solution

This does add @babel/runtime as a dependency to ensure it's on the node path in case they are using es modules. For non-es builds it will simply be ignored so no problem. Most of the time if they are also using @babel/runtime the packager will hoist them to the same version if they can match. Worst case scenario is there are two versions which degrades bundle size to where it was previously.

# TODO

- [ ] Add & update tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "@babel/core": "^7.0.0",
     "@babel/plugin-proposal-class-properties": "^7.0.0",
     "@babel/plugin-proposal-object-rest-spread": "^7.0.0",
+    "@babel/plugin-transform-runtime": "^7.5.5",
     "@babel/preset-env": "^7.0.0",
     "@babel/preset-flow": "^7.0.0",
     "babel-core": "^7.0.0-bridge.0",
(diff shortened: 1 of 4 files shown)


paularmstrong:
Can you run a build before & after? Copy the output and paste it up in a gist so we can see what the full savings are?

I've also considered getting rid of rollup completely and just using babel to compile commonjs and es modules.

package.json
```diff
@@ -77,9 +78,13 @@
...
     "rollup-plugin-babel": "^4.3.2",
+    "rollup-plugin-commonjs": "^10.1.0",
     "rollup-plugin-filesize": "^6.1.1",
+    "rollup-plugin-node-resolve": "^5.2.0",
     "rollup-plugin-terser": "^5.0.0",
     "typescript": "^3.4.5"
   },
-  "dependencies": {}
+  "dependencies": {
+    "@babel/runtime": "^7.5.5"
```
paularmstrong: Can this be a peer dependency instead?

paularmstrong (changes requested):
ping @ntucker still waiting on the comparison data

PR #335, opened 2018-05-10
Add support for circular references

ptshih:
# Problem

Unable to normalize inputs with circular references. I get a `maximum call stack size exceeded` error.

# Solution

This PR is based on https://github.com/paularmstrong/normalizr/pull/330 with a few changes (based on the fixes recommended there):

- Used `Array.prototype.some` for equality check instead of `Set` to avoid a polyfill
- Added tests using `toMatchSnapshot`

# TODO

- [x] Add & update tests
- [x] Ensure CI is passing (lint, tests, flow)
- [ ] Update relevant documentation

PR #420, opened 2019-09-28
Improve performance of circular reference detection

wurstbonbon:
# Problem

The support for circular references added with #335 comes with a significant performance cost.
For every entity the list of already parsed entities is searched. This leads to non-linear performance characteristics.

# Solution

I’ve partitioned the visited entities lookup by ids. Entities with a different type that share the same id will still be added to a list, which shouldn't be a problem in practice. Using a Set would have been simpler, but that was ruled out [earlier](https://github.com/paularmstrong/normalizr/pull/330#discussion_r179919159).

### Setup

```
const data = [];
for (let i = 0; i < N; i++) {
  data.push({ id: i, type: `foo-${i}` });
}
normalize(data, [mySchema]);
```

### Normalization times in milliseconds for 5 runs

|N = 10,000 entities   |  |  |  |  |          |
| -------- | ---:| ---:| ---:| ---:| ---:|
|current      |157|141|139|148|142|
|optimized | 38|  37|  39| 38|  36|

|N = 100,000 entities  |  |  |  |  |                   |
| -------- | ----:| ----:| ----:| ----:| ----:|
|current      |6696|6738|6762|6772|6673|
|optimized |  255|   248|    259|   257|   241|

# TODO

- [ ] Add & update tests

### DIFF
--- a/src/schemas/Entity.js
+++ b/src/schemas/Entity.js
@@ -49,10 +49,15 @@ export default class EntitySchema {
   }
 
   normalize(input, parent, key, visit, addEntity, visitedEntities) {
-    if (visitedEntities.some((entity) => entity === input)) {
-      return this.getId(input, parent, key);
+    const id = this.getId(input, parent, key);
+
+    if (!(id in visitedEntities)) {
+      visitedEntities[id] = [];
     }
-    visitedEntities.push(input);
+    if (visitedEntities[id].some((entity) => entity === input)) {
+      return id;
+    }
+    visitedEntities[id].push(input);
 
     const processedEntity = this._processStrategy(input, parent, key);
     Object.keys(this.schema).forEach((key) => {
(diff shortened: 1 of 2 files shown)


src/schemas/Entity.js
```diff
@@ -49,10 +49,15 @@ export default class EntitySchema {
...
+    const id = this.getId(input, parent, key);
+
+    if (!(id in visitedEntities)) {
+      visitedEntities[id] = [];
+    }
+    if (visitedEntities[id].some((entity) => entity === input)) {
+      return id;
     }
-    visitedEntities.push(input);
+    visitedEntities[id].push(input);
```
paularmstrong: Just another implementation thought, no need to act on it:

I wonder if `visitedEntities` could be a map structured by key, then id:
```js
visitedEntities = {
  [key: string]: { [id: string]: entity }
};
```

As it is now, entities of the same type are jumbled together if they have the same `id`, which kind of just makes it weird if someone were to step through and inspect it with their debugger.

wurstbonbon: Yes, I skipped the `key` for compactness, but you’re right that would be more explicit. I can make that change if you want.

wurstbonbon: I just did it, I hope this can be merged soon now. This causes a big performance hit for us.

Issue #421, opened 2019-10-01
A fallback strategy for denormalization for missing entities


bjartebore:
# Problem

In some cases I would like to denormalize a structure with a list of Ids instead of a list of subobjects (reasons could be to limit fetched data). When i normalize this structure the normalization works, I am left with the entity ids, but there is no way to create stubs for the entities itself, which I find ok since there could be an instance of the same object somewhere else in the structure.

But when denormalizing this I will get back an array with undefined items. 

# Solution

A possible solution to this could be to add a possibility to add a fallbackStrategy option to the EntitySchema where we could write our own fallback method for denormalization.

I would be glad to create a pull request for this.

PR #422, opened 2019-10-01
added a fallback strategy for denormalization

bjartebore:
# Problem

In some cases I would like to denormalize a structure with a list of Ids instead of a list of subobjects (reasons could be to limit fetched data). When i normalize this structure the normalization works, I am left with the entity ids, but there is no way to create stubs for the entities itself, which I find ok since there could be an instance of the same object somewhere else in the structure.

But when denormalizing this I will get back an array with undefined items. 

# Solution

A possible solution to this could be to add a possibility to add a fallbackStrategy option to the EntitySchema where we could write our own fallback method for denormalization.
# TODO

- [x] Add & update tests

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -175,6 +175,9 @@ const normalizedData = normalize(data, myArray);
     - `value`: The input value of the entity.
     - `parent`: The parent object of the input array.
     - `key`: The key at which the input array appears on the parent object.
+  - `fallbackStrategy(key, schema)`: Strategy to use when denormalizing data structures with id references to missing entities.
+    - `key`: The key at which the input array appears on the parent object.
+    - `schema`: The schema of the missing entity
 
 #### Instance Methods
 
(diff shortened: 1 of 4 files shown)


bjartebore:
#421

bjartebore:
I have updated the relevant documentation, I hope it will suffice :)

mecp:
Hi guys, Is there a way to skip missing references while denormalizing to avoid having `[undefined, undefined, undefined]` situation?

for example here, in a use case I have, I want denormalizer  to return `c: []` instead of `c: [undefined, undefined, undefined]`

```ts
const { schema, denormalize } = require('normalizr');

const a = {
  id: 1,
  b: 1,
  c: [1, 2, 3]
}

const cSchema = new schema.Entity('c');
const aSchema = new schema.Entity('a', { c: [cSchema] });

const result = denormalize([1], [aSchema], { a: { 1: a }, c: {} })
```
available to run here https://runkit.com/mecp/5eb5baa8aad698001bb12788

PR #439, opened 2020-04-07
fix example of fallbackStrategy

aimergenge:
# Problem

The example of fallbackStrategy is broken and missing important part of code.

# Solution

See the code.

# TODO

- [x] Add & update tests

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -258,15 +258,15 @@ normalize(data, [patronsSchema]);
 
 #### `fallbackStrategy` Usage
 ```js
-const users = [
-  { id: '1', name: "Emily", requestState: 'SUCCEEDED' },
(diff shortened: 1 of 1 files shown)


paularmstrong:
> broken and missing important part of code.

Can you explain what it means to be "broken"? What is it missing?

docs/api.md
```diff
@@ -280,6 +277,23 @@ const bookSchema = new schema.Entity('books', {
...
+})
+
+function toEntity(arr, idAttribute = 'id') {
+  return arr.reduce((acc, x) => {
+    return {
+      ...acc,
+      [x[idAttribute]]: x
+    }
+  }, {})
+}
```
paularmstrong, line 296: Helpers like this in examples cause confusion and should be avoided.

PR #453, opened 2020-08-18
user entity updated to author for more clarity

faizanu94:
# Problem

In [blog post](https://github.com/paularmstrong/normalizr#quick-start), the `user` entity not used in sample JSON, however used in the `normalizr` code example

# Solution

The `user` entity is changed to `author` for more clarity

# TODO

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,22 +61,22 @@ Consider a typical blog post. The API response for a single post might look some
 }
 ```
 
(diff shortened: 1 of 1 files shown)


paularmstrong:
@luccamordente is right, the original example is correct as both `author` and `commenter` are both the same `user` entity, just attached to different objects

PR #461, opened 2020-12-09
fix(array): When An Array Denormalize cannot find an Entity, don't say it is an array item

collides:
# Problem

When I use Array denormalize, if I can't find an Entity: 

```typescript

const ids = [1,2,6]

const entities = [
    {
       id: 1,
       name: 'test-1'
    },
    {
      id: 2,
      name: 'test-2'
    }
]

const exampleDenormalize = denormalize(ids, schema, entities)

// [{id: 1, name: ..,}, {id: 2, name: ...}, undefined]

```

It's not easy to speculate about arrays, whether they exist or not

Obviously I need to add filter to all of the Denormalizer

# Solution

Using filters in an Array denormalize()  solves this problem uniformly

# TODO

- [x] Add & update tests

### DIFF
--- a/src/schemas/Array.js
+++ b/src/schemas/Array.js
@@ -36,6 +36,8 @@ export default class ArraySchema extends PolymorphicSchema {
   }
 
   denormalize(input, unvisit) {
-    return input && input.map ? input.map((value) => this.denormalizeValue(value, unvisit)) : input;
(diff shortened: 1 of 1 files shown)


paularmstrong:
The current behavior is intended to match the output with your input.

PR #462, opened 2020-12-12
schemaAttribute

exekiel87:
# Problem

Using schemaAttribute on Array or Values don't promotes code reuse.

# Solution

Implementation omitted on readme and examples added to use Union instead.


acfasj:
this PR helps me understand UnionSchema better

PR #474, opened 2021-05-18
Fix incorrect typescript definition file

### LINKED ISSUE #470: Typescript types included with normalizr are incorrect
# Problem

Because of the way that schemas are defined and typed, the output type of `normalize` does not match the actual entity types whenever you have nested schemas (which should be most use cases), which can cause issues anywhere the result is read.

**Input**

```ts
interface Player {
    id: string
    characters: string[]
}

interface Character {
    id: string
    name: string
}

type Entity<T> = { [key: string] : T } | undefined

const characterSchema = new schema.Entity<Character>('characters'); // Schema.Entity<Character>
const playerSchema = new schema.Entity<Player>(
    'players',
    { 'characters': characterSchema }
); // Schema.Entity<Player> <-- no reference to characters!

const myData = [
    {
        'id': 1,
        'characters': [{'id': 1, 'name': 'Bill'}, {'id': 2, 'name': 'Greg']
    },
    {
        'characters': [{'id': 3, 'name': 'Sue'}]
    }
];

const result = normalize(myData, playerSchema);

const players: Entity<Player> = result.entities.players;

/**
  * This throws an error: 
  * Type '{ [key: string]: Player; } | undefined' is not assignable to type 'Entity<Character>'.
  *  Type '{ [key: string]: Player; }' is not assignable to type 'Entity<Character>'.
  *    Index signatures are incompatible.
  *      Property 'name' is missing in type 'Player' but required in type 'Character'.
  */
const characters: Entity<Character> = result.entities.characters;
```

**Output**

* The output type of normalize should include BOTH `Player` and `Character` in some fashion. 
```
normalize(myData, playerSchema): NormalizedSchema<E = { 'players': { [key: string]: Player } | undefined, 'characters': { [key: string]: Character } | undefined }, R = string[]>
```

Here's what I _actually_ see when I run the above:

```
normalize(myData, playerSchema): NormalizedSchema<E = { [key: string]: { [key: string]: Player } | undefined}, R = any>
```

grandivory:
# Problem

Fixes https://github.com/paularmstrong/normalizr/issues/470

# Solution
I ended up having to touch a lot more files in the project than I expected to. A summary of changes is below:

* Rewrote the index.d.ts file from scratch to include as much type information as possible given the structure of the project.
* Moved index.d.ts next to index.js
  * This allows typescript tests to read the types properly within the project
  * Updated the build step to include the .d.ts file in dist, and updated the package.json definition with the new location of the types file
* Updated the version of typescript included in the project to latest
  * Added a new `tsconfig.json` file
* Reworked typescript tests - Running `yarn test` will now verify not only that the code is functional, but also that the type definitions work
  * Moved the compile-only tests into a `type-tests` subdirectory.
  * Added new compile-only tests for both positive and negative cases
  * Added a new tool, `check-dts`, that will check that test files compile, or that they throw particular errors
  * Duplicated the existing javascript tests to typescript
    * Updated the jest config to run typescript tests as well
    * Added a rule to babel to be able to process the typescript files and updated the babel parser
* Added the `cross-env` package to allow for development on windows

# TODO

- [x] Add & update tests


typescript-tests/Entity.test.ts
```diff
@@ -0,0 +1,337 @@
+import { denormalize, IdFunction, normalize, schema } from '../src';
```
paularmstrong, line 1: Why are all of these test files complete duplicates of the source test files? I would prefer to not need to maintain duplicates like this.

grandivory: Since the test files verify that all of the functionality works as expected, duplicating the files and having the typescript compiler run on them was the simplest way to verify that all of the functionality works as expected within typescript as well. There was actually one test that did have to change, in Array.test.ts `denormalizes multiple entities`: The entry of `{ id: { id: '789' ), schema: 'dogs' }` wasn't valid given the types that I assigned, and it didn't look like it actually should be, based on the usage of denormalize.

Another option would be to remove the javascript tests, leaving only the typescript ones for most functionality. That would test both that the functionality works and that the given type definitions are correct. The only things that would need to be added at that point are tests that expect to throw an error in javascript, but wouldn't compile in typescript (e.g. passing `42` to the entity key).

I'm happy to make that change if you'd like; I just didn't want to remove existing tests on my own.

grandivory: @paularmstrong I've removed the duplicate tests. This also has the advantage that you can now clearly see what's possible in javascript but not in typescript with the new definitions file.

adesege:
Any update on this? This is blocking us

x4th:
Can you please take a look at this again? @paularmstrong

f-g-p:
Hey, This would be quite useful for us too :)

Thanks for the work!

